PR #2, opened 2018-08-13
Added Symfony 3.4.14 (w/ flex) + server

stephenwf:


thewilkybarkid:
We should use Symfony 4 (see #3). Sorry for duplicated effort!

PR #4, opened 2018-08-14
Set up Docker

thewilkybarkid:


src/Kernel.php
```diff
@@ -13,6 +13,11 @@
 {
     use MicroKernelTrait;
 
+    public function getProjectDir()
+    {
+        return __DIR__.'/..';
```
thewilkybarkid: Symfony guesses by finding a `composer.json`. Since it's not in the `fpm` image, we have to be explicit.

docker-compose.override.yaml
```diff
@@ -0,0 +1,17 @@
...
+        volumes:
+            - ./composer.json:/app/composer.json
+            - ./composer.lock:/app/composer.lock
+            - ./symfony.lock:/app/symfony.lock
+            - vendor:/app/vendor
+    fpm:
+        volumes:
+            - ./:/var/www/html/
+            - /var/www/html/var
+            - vendor:/var/www/html/vendor
```
thewilkybarkid: Usual caveat that this will be shared between builds. Would be interested to know if others have solutions. /cc @elrossco22 @stephenwf

docker-compose.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+services:
+    composer:
+        build:
+            context: .
+            dockerfile: Dockerfile.composer
+            args:
+                composer_dev_arg: ${COMPOSER_DEV_ARG}
+        image: libero/browser_composer:${IMAGE_TAG}
+        command: 'true'
```
thewilkybarkid: Never found a way to have `docker-compose` services that don't actually run.

Dockerfile
```diff
@@ -0,0 +1,14 @@
+ARG image_tag=latest
+
+FROM libero/browser_composer:${image_tag} AS composer
+FROM php:7.2.8-fpm-alpine3.7
```
giorgiosironi: the base images sound good to me but agreed that they should be the subject of wider discussion @stephenwf @elrossco22

src/Kernel.php
giorgiosironi: It could be useful to have the `composer.lock` file in the image to check which package versions are available, but `vendor/composer/installed.json` already has all the information to track this

Dockerfile
```diff
@@ -0,0 +1,14 @@
+ARG image_tag=latest
```
giorgiosironi: a branching model could influence this, but as long as there is only `master` it makes sense for the default tag to be `latest` (which isn't necessarily pushed when built)

Dockerfile.composer
```diff
@@ -0,0 +1,14 @@
...
+ARG composer_dev_arg
+
+COPY composer.json \
+    composer.lock \
+    symfony.lock \
+    ./
+
+RUN composer --no-interaction install ${composer_dev_arg} --ignore-platform-reqs --no-autoloader --no-suggest --prefer-dist
+
+COPY src/ src/
```
giorgiosironi: [I heard](https://github.com/symfony/recipes/blob/master/README.rst#copy-from-recipe-configurator) Symfony Flex could also write to `config/`?

giorgiosironi: (but maybe not during these `composer` commands)

thewilkybarkid: Yeah, when adding a new dependency. Might need to mount `config/` to `composer` so that changes are captured if using Composer through the container.

docker-compose.override.yaml
```diff
@@ -0,0 +1,17 @@
+version: '3'
+
+services:
+    composer:
+        volumes:
+            - ./composer.json:/app/composer.json
+            - ./composer.lock:/app/composer.lock
+            - ./symfony.lock:/app/symfony.lock
```
giorgiosironi: had some success in using `pipenv` through the related image and mounted volumes, to use a consistent version rather than relying on the host's one. Could be done on this if it mounted the folders Symfony Flex writes to?

Dockerfile.composer
thewilkybarkid: Added in f853fa9afd17bbe8faf8b0ac517dcd9709640fba (and tested by running `docker-compose run composer composer require orm`).

docker-compose.yaml
giorgiosironi: `/bin/true` could be immediately understood

Dockerfile.composer
```diff
@@ -0,0 +1,14 @@
+FROM composer:1.7.1
+
+ARG composer_dev_arg
+
+COPY composer.json \
+    composer.lock \
+    symfony.lock \
+    ./
+
+RUN composer --no-interaction install ${composer_dev_arg} --ignore-platform-reqs --no-autoloader --no-suggest --prefer-dist
```
stephenwf: Is there value in adding `--no-scripts` ?

thewilkybarkid: Not now (since there are no scripts). We've run into problems in the past with this too (https://github.com/composer/composer/issues/7100).

.env.dist
```diff
@@ -1,3 +1,6 @@
+COMPOSER_DEV_ARG=
```
giorgiosironi: this is named `.env.dist` so it won't be read by `docker-compose` (or is it?)

Is it meant to be copied onto `.env` for customization? If so `.env` should be ignored?

giorgiosironi: It is ignored https://github.com/libero/browser/blob/master/.gitignore#L2

giorgiosironi: Although this file is overloaded: it contains partly variables for `docker-compose.yml` files and partly variables for the containers in it to use

thewilkybarkid: Thinking about it, running `composer install` _outside_ saw `.env` created for me (based on `.env.dist`). Must be some Symfony Flex step. Taking a look.

thewilkybarkid: We can set defaults in `docker-compose.yaml` (means repeating it though).

giorgiosironi: We can override the file's values if need be with `source some-file.env && docker-compose up`, but no problem in using a single file if there are no variables named the same way and they are separated in two sections like in this example.

thewilkybarkid: Will add a comment to say these are Docker-Composer envvars (if you're using the container elsewhere you won't need these ones).

docker-compose.yaml
```diff
@@ -0,0 +1,27 @@
...
+            args:
+                composer_dev_arg: ${COMPOSER_DEV_ARG}
+        image: libero/browser_composer:${IMAGE_TAG}
+        command: /bin/true
+    fpm:
+        build:
+            context: .
+            args:
+                image_tag: ${IMAGE_TAG}
+        image: libero/browser:${IMAGE_TAG}
```
stephenwf: Should we parameterise the image name too?

docker-compose.override.yaml
stephenwf: Sorry @thewilkybarkid not sure I follow, shared between builds where?

docker-compose.yaml
thewilkybarkid: For creating `libero/browser_unstable:12345` etc? Can do if needed (maybe `${IMAGE_SUFFIX}`?), can this be done by whatever's going to create that image?

docker-compose.override.yaml
thewilkybarkid: Been using this technique to 'unmount' subdirectories  (so `var` isn't pulled back to the host, and `vendor` comes from the `composer` image rather than the host (which could be empty, so would delete the contents)). The problem is that the volume is persisted, so changes might not be recognised on future builds. See also https://github.com/libero/walking-skeleton/pull/34.

If we're going Docker only, we might be able to stop these being subdirectories by having them elsewhere on the filesystem (means not using the standard Composer `vendor` directory etc).

docker-compose.yaml
giorgiosironi: The build can pass in the necessary `IMAGE_SUFFIX`, but it needs a parameter to be added here. Not a blocker.

Dockerfile
giorgiosironi: Poll on Slack shows preference for these base images so far wrt Fedora and larger containers with nginx+fpm.

giorgiosironi:
Imho deferred until we need a proper production deployment (the base images would need to change)

giorgiosironi:
Some defaults are needed to prevent this failure:
```
$ docker-compose build
WARNING: The COMPOSER_DEV_ARG variable is not set. Defaulting to a blank string.
WARNING: The IMAGE_TAG variable is not set. Defaulting to a blank string.
Building composer
ERROR: invalid reference format
```

docker-compose.override.yaml
```diff
@@ -0,0 +1,18 @@
+version: '3'
+
+services:
+    composer:
+        volumes:
+            - ./config/:/app/config/
```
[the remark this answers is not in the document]
thewilkybarkid: It's so config created by Symfony Flex when installing a new dependency is captured. Though come to think of it, it's wider than just this directory (`.gitignore`, for example).

PR #5, opened 2018-08-16
Add simple PHPUnit test

thewilkybarkid:


.gitignore
```diff
@@ -1,3 +1,7 @@
+###> phpunit/phpunit ###
```
giorgiosironi: Is the `### ... ###` wrapping here also written by Flex? Why is it interested in PHPUnit rather than Symfony packages?

.travis.yml
```diff
@@ -5,13 +5,14 @@ language: minimal
 services: docker
 
 install:
-    - docker-compose build
+    - docker-compose -f docker-compose.yaml build
```
giorgiosironi: Github search suggests the `docker-compose.yml` naming is more common https://github.com/search?q=docker-compose.yml

.gitignore
thewilkybarkid: Yep. It can work with any vendor.

composer.lock
```diff
@@ -1185,7 +1185,1527 @@
...
+                }
+            },
+            "autoload": {
+                "classmap": [
+                    "src/"
+                ]
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "BSD-3-Clause"
```
giorgiosironi: not a problem specific to here, but do we need to check something about the dependencies licenses compatibility? Not for now but as soon as we distribute Docker images

tests/WebTest.php
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * @test
+     */
+    public function it_returns_a_404_when_previewing_a_404_page() : void
+    {
+        $client = static::createClient();
+
+        $client->request('GET', '/');
+
+        $this->assertSame(404, $client->getResponse()->getStatusCode());
```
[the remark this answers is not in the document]
thewilkybarkid: This talks directly to the app kernel.

composer.lock
thewilkybarkid: Seems to do the job. Currently everything in `require` is MIT (this example is a dev dependency).

tests/WebTest.php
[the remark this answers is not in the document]
giorgiosironi: (as in, a single PHP process) :+1:

docker-compose.yaml
```diff
@@ -25,3 +25,12 @@ services:
             - 8080:80
         depends_on:
             - fpm
+    ci:
```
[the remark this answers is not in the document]
thewilkybarkid: The default is `--dev`. IIRC you get a deprecation notice setting it explicitly.

PR #7, opened 2018-08-17
Add basic content page

thewilkybarkid:


config/routes.yaml
```diff
@@ -0,0 +1,5 @@
+content:
+    path: /content/{id}
+    controller: Libero\ContentPageBundle\Controller\ContentController
```
thewilkybarkid: Currently allows for one instance, and the route has to be set manually. Will need to make bundle configuration to set up multiple.

vendor-extra/ContentPageBundle/src/Controller/ContentController.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Libero\ContentPageBundle\Controller;
+
+use Symfony\Component\HttpFoundation\Response;
+
+final class ContentController
+{
+    public function __invoke(string $id) : Response
+    {
+        return new Response($id, Response::HTTP_OK, ['Content-Type' => 'text/plain']);
```
thewilkybarkid: Next step is to use a client to request a content `front` (which should just be static/faked data for now).

config/routes.yaml
```diff
@@ -0,0 +1,11 @@
...
+    path: /articles/{id}
+    controller: libero.content_page.controller.content.scholarly_article
+    requirements:
+        id: '[A-Za-z0-9\-._]+'
+
+blog-article:
+    path: /blog/{id}
+    controller: libero.content_page.controller.content.blog_article
+    requirements:
+        id: '[A-Za-z0-9\-._]+'
```
thewilkybarkid: These should probably be enabled automatically based on configuration, otherwise has the same problem as https://github.com/libero/browser/pull/7#r210852139.

PR #9, opened 2018-08-17
Alternative to composer-merge plugin

stephenwf:
Posting this as an alternative to using composer merge. Aside from not using the merge-logic itself, it also forces `browser` itself define the dependencies it has (and omit those that it doesn't). When it comes to removing the dependency into its own file its just a case of removing it form the filesystem and `composer update`. 

One last point, with the library approach, its likely we will have Bundles (in `/vendor-extra`) that will depend on these libraries (that also live in `/vendor-extra`) at which point, if we don't define them in the `composer.json` file, will be harder to track down later. This is more of a vanilla composer approach.

Bonus to this approach, when we do come to splitting these out into repositories, you can continue to develop them simply by cloning into the `/vendor-extra` locally and it will pick up changes.

### DIFF
--- a/Dockerfile.composer
+++ b/Dockerfile.composer
@@ -7,9 +7,10 @@ COPY composer.json \
     symfony.lock \
     ./
 
+COPY vendor-extra/ vendor-extra/
+
 RUN composer --no-interaction install ${composer_dev_arg} --ignore-platform-reqs --no-autoloader --no-suggest --prefer-dist
 
 COPY src/ src/
-COPY vendor-extra/ vendor-extra/
 
 RUN composer --no-interaction dump-autoload ${composer_dev_arg} --classmap-authoritative
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
         "php": "^7.2",
         "ext-ctype": "*",
         "ext-iconv": "*",
+        "libero/content-page-bundle": "0.0.1",
         "symfony/config": "^4.1",
         "symfony/console": "^4.1",
         "symfony/dependency-injection": "^4.1",
@@ -23,8 +24,7 @@
         "symfony/http-foundation": "^4.1",
         "symfony/http-kernel": "^4.1",
         "symfony/routing": "^4.1",
-        "symfony/yaml": "^4.1",
-        "wikimedia/composer-merge-plugin": "^1.4"
+        "symfony/yaml": "^4.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^7.3",
@@ -38,13 +38,15 @@
         "symfony/polyfill-php70": "*",
         "symfony/polyfill-php71": "*"
     },
-    "extra": {
-        "merge-plugin": {
-            "include": [
-                "vendor-extra/*/composer.json"
-            ]
-        }
+    "scripts": {
+        "test": "phpunit"
     },
+    "repositories": [
+        {
+            "type": "path",
+            "url": "./vendor-extra/*"
+        }
+    ],
     "config": {
         "platform": {
             "php": "7.2.8"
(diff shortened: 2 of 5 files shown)


giorgiosironi:
The trick is that the autoload generation picks up `vendor-extra` I guess, so you can put anything in there. This seems cleaner to me than a Composer plugin, but the plugin may be doing something I don't know like running hooks of some kind.

Dockerfile.composer
```diff
@@ -7,9 +7,10 @@ COPY composer.json \
     symfony.lock \
     ./
 
+COPY vendor-extra/ vendor-extra/
+
 RUN composer --no-interaction install ${composer_dev_arg} --ignore-platform-reqs --no-autoloader --no-suggest --prefer-dist
 
 COPY src/ src/
-COPY vendor-extra/ vendor-extra/
```
thewilkybarkid: Increase cache busting is a downside. (Though as Flex downloads in parallel it's not that slow.)

vendor-extra/ContentPageBundle/composer.json
```diff
@@ -2,6 +2,7 @@
     "name": "libero/content-page-bundle",
     "description": "A Libero content page",
     "type": "symfony-bundle",
+    "version": "0.0.1",
```
thewilkybarkid: IIRC the downside of using `path` repos is it not recognising updates. Guess this gets round it? Slight annoyance having to bump on every change.

composer.lock
```diff
@@ -1,11 +1,41 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
```
thewilkybarkid: Update your local Composer, or, ideally, use the container.

composer.json
```diff
@@ -38,13 +38,15 @@
...
         "symfony/polyfill-php71": "*"
     },
-    "extra": {
-        "merge-plugin": {
-            "include": [
-                "vendor-extra/*/composer.json"
-            ]
-        }
+    "scripts": {
+        "test": "phpunit"
```
thewilkybarkid: Use the container instead.

vendor-extra/ContentPageBundle/composer.json
stephenwf: It symlinks the whole thing, so the version only needs to be bumped once published. At that point though, it probably would be in its own repository (and not need this field)

Dockerfile.composer
stephenwf: We could volume the composer cache directory if it becomes a problem I guess

stephenwf:
> the plugin may be doing something I don't know like running hooks of some kind.
@giorgiosironi If the plugin is doing this, then that will break when we move them to packagist. From the composer docs:
> Note: Only scripts defined in the root package's composer.json are executed. If a dependency of the root package specifies its own scripts, Composer does not execute those additional scripts.

composer.lock
```diff
@@ -4,8 +4,43 @@
...
-    "content-hash": "ff2c1392af4a32ee1bda12bf7c90a4ce",
+    "content-hash": "df773ab4a17b82868a21f8a1b87d49cf",
     "packages": [
+        {
+            "name": "libero/content-page-bundle",
+            "version": "dev-master",
+            "dist": {
+                "type": "path",
+                "url": "./vendor-extra/ContentPageBundle",
+                "reference": "a334e0f10e447c85dbc6a543dc9d0864a458139a",
```
thewilkybarkid: @stephenwf Looked at not having to set the `version` manually, looks like this is a hash of the bundle contents so is enough?

stephenwf: oh nice one!

PR #10, opened 2018-08-20
[WIP] Create api client

jkopanas:


vendor-extra/ApiClientBundle/src/data/body.xml
```diff
@@ -0,0 +1,643 @@
+<?xml version="1.0" ?>
```
thewilkybarkid: Right now we only need to worry about the `front`.

vendor-extra/ApiClientBundle/src/data/front.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" ?>
+<?xml-model href="http://localhost:8081/schemas/ijm/article/front.rng"?>
+
+<front xmlns="http://libero.pub" xmlns:ms="http://microsimulation.org" xml:lang="en">
+
+    <id>IJM_2017_10_2_2</id>
+    <title>Coordination in Centralized and Decentralized Systems</title>
+    <abstract id="abstract">
```
thewilkybarkid: We only need the `<id>` and `<title>` for now.

vendor-extra/ContentPageBundle/src/DependencyInjection/ContentPageExtension.php
```diff
@@ -26,9 +26,9 @@ private function addPage(string $name, array $config, ContainerBuilder $containe
     {
         $id = sprintf(self::CONTENT_CONTROLLER_ID, $name);
         $definition = new Definition(ContentController::class);
-
+       
+        $definition->setAutowired(true);
```
thewilkybarkid: Autowiring's discourages for bundles. We'll need to make the client configurable (will most likely be one content page per content store, which has one client).

vendor-extra/ApiClientBundle/src/ApiClientInterface.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Libero\ApiClientBundle;
+
+use GuzzleHttp\Promise\PromiseInterface;
+use Psr\Http\Message\RequestInterface;
+
+interface ApiClientInterface
+{
+    public function getData();
```
thewilkybarkid: We should let this return a promise of the response.

vendor-extra/ApiClientBundle/src/Services/Article.php
```diff
@@ -0,0 +1,22 @@
...
+namespace Libero\ApiClientBundle\Services;
+
+use Libero\ApiClientBundle\ApiClientInterface;
+use GuzzleHttp\Client;
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\Promise\PromiseInterface;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+final class Article implements ApiClientInterface
```
thewilkybarkid: We've switched from using the term `Article` to the more generic `Content`. This should be called `ContentStoreClient`, and provide ways to access each of the content store API endpoints and methods (being worked on in https://github.com/libero/schemas/pull/1/, described in RAML).

vendor-extra/ApiClientBundle/src/Services/GuzzleXml.php
```diff
@@ -0,0 +1,26 @@
+<?php
+namespace Libero\ApiClient\Services;
+
+use GuzzleHttp\Client;
+use GuzzleHttp\Promise\PromiseInterface;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+final class GuzzleXml
```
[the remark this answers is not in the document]
thewilkybarkid: Though maybe the dummy adapter would be better at the `ContentStoreClient` level (see https://github.com/libero/browser/pull/10#discussion_r211281801).

vendor-extra/ApiClientBundle/src/ApiClientInterface.php
thewilkybarkid: So refers to make a HTTP request to a Libero API.

vendor-extra/ApiClientBundle/src/Services/Article.php
thewilkybarkid: This can be done when #13 is merged.

PR #11, opened 2018-08-21
Push Docker image

giorgiosironi:
Needs to be on an origin branch rather than a fork to access environment variables for Docker Hub access.

Should normally only be executed on a tag or on `master`.

Without a Docker Hub organization it's too early to differentiate between stable and unstable, so this will be an image with a certain tag that will be deployed to the `unstable` environment as an additional step when it becomes available.


.travis.yml
```diff
@@ -4,6 +4,13 @@ language: minimal
 
 services: docker
 
+branches:
+    only:
+        - master
```
giorgiosironi: better way to instrument Travis CI than the point and click interface, allows the mainline and PRs to be built but not all pushed branches

.docker/push-image.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+# should be executed only if $TRAVIS_BRANCH = master
```
giorgiosironi: continuing tomorrow

.docker/push-image.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
thewilkybarkid: Might be better to put `.travis`.

.travis.yml
```diff
@@ -17,3 +24,4 @@ script:
 
 after_script:
     - docker-compose -f docker-compose.yaml down
+    - .docker/push-image.sh
```
thewilkybarkid: Will push after failures too. Probably ok.

giorgiosironi: That's _really_ unstable, will move to `after_success`

.travis.yml
```diff
@@ -17,3 +24,6 @@ script:
 
 after_script:
     - docker-compose -f docker-compose.yaml down
+
+after_success:
```
thewilkybarkid: Looks like it's [run before `after_script`](https://docs.travis-ci.com/user/customizing-the-build/#the-build-lifecycle), would be useful to keep the same order for clarity.

.travis/push-image.sh
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+if [[ "$TRAVIS_BRANCH" = "master" && "$TRAVIS_PULL_REQUEST" = "false" ]]; then
+    echo "$DOCKER_PASSWORD" | docker login --username "$DOCKER_USERNAME" --password-stdin 
+
+    # tag temporarily as liberoadmin due to lack of `libero/` availability
+    docker tag "libero/browser:$IMAGE_TAG" "liberoadmin/browser:$IMAGE_TAG"
+    docker push "liberoadmin/browser:$IMAGE_TAG"
+else
+    echo "not pushing an image, we are not on master"
```
thewilkybarkid: Fine for now. We can improve messages when we update to use the proper organisation and handle tags.

giorgiosironi:
First run in https://travis-ci.com/libero/browser/builds/82548926

PR #14, opened 2018-09-18
Trigger downstream environment repository build

giorgiosironi:
The `environments` repository deploys the latest version of each project to the `unstable` environment. It will be theoretically possible to specify a particular Docker image tag to deploy, but left for future development.

The build being triggered looks like this:

(already added the TRAVIS_TOKEN environment variable)


.travis.yml
```diff
@@ -27,3 +27,9 @@ after_success:
 
 after_script:
     - docker-compose -f docker-compose.yaml down
+
+deploy:
+    provider: script
+    script: .travis/downstream.sh $TRAVIS_COMMIT
```
thewilkybarkid: `downstream` is a bit too generic, might end up with multiple downstreams.

.travis/downstream.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -e
+
+body='{
+	"request": {
```
thewilkybarkid: Spaces

.travis/downstream-environments.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -e
+
+body='{
+    "request": {
+        "branch":"master",
+        "message":"browser new version: '"$1"'"
```
thewilkybarkid: Is this both human-readable and parsable? Could be `Update Browser to 12345`?

giorgiosironi: Human-oriented, will update

PR #18, opened 2018-11-19
Get content title from the API

thewilkybarkid:


vendor-extra/ContentPageBundle/tests/GuzzleTestCase.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public function setUpGuzzle() : void
+    {
+        $this->filesystem = new FileSystem();
+
+        $stack = HandlerStack::create(new MockHandler());
+        $stack->push(
+            new MockMiddleware(
+                $this->mock = new MockStorageAdapter(
+                    $this->filesystem->path(''),
```
thewilkybarkid: Easier than https://github.com/elifesciences/api-sdk-php/blob/master/test/InMemoryStorageAdapter.php.

.docker/nginx/browser.conf
```diff
@@ -17,8 +17,8 @@ server {
         fastcgi_pass app:9000;
         fastcgi_split_path_info ^(.+\.php)(/.*)$;
 
-        fastcgi_param SCRIPT_FILENAME $realpath_root$fastcgi_script_name;
-        fastcgi_param DOCUMENT_ROOT $realpath_root;
+        fastcgi_param SCRIPT_FILENAME /app/public/$fastcgi_script_name;
```
giorgiosironi: doesn't match `root`?

api/data/scholarly-articles/article1/1.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
giorgiosironi: all the files in `api/` seem to be for a testing/development configuration of the `dummy-api` container, hence any way to make that more clear? `.api` hidden folder or `.docker` or move in `tests/` or some other subfolder that is clearly not in the application?

vendor-extra/ContentPageBundle/tests/GuzzleTestCase.php
giorgiosironi: It seems easier to setup. What is the empty path?

composer.json
```diff
@@ -28,13 +29,17 @@
         "symfony/yaml": "^4.1"
     },
     "require-dev": {
+        "csa/guzzle-cache-middleware": "^1.0",
```
giorgiosironi: weird that this is actually used for mocking, rather than for caching

giorgiosironi: but `require-dev` is the right place then

.docker/nginx/browser.conf
giorgiosironi: I see the folders are mounted in different places in the `nginx` and the other app containers, though this shows how to make the paths different if needed, it could be easier to keep everything consistent for maintenance

giorgiosironi: Now supporting this, with named volumes in `sample-configuration`

vendor-extra/ContentPageBundle/tests/GuzzleTestCase.php
thewilkybarkid: To get the absolute root path. Might be a more obvious way...

api/data/scholarly-articles/article1/1.xml
thewilkybarkid: `.docker` sounds like the best place (used in `dev` only).

composer.json
thewilkybarkid: It's annoying actually a hard dependency of `csa/guzzle-bundle`, but we only use it here.

README.md
```diff
@@ -6,15 +6,19 @@ Libero browser
...
+To run a website reading from two content services (`blog-articles` and `scholarly-articles`):
+
 1. Create `config/packages/content_page.yaml`:
 
     ```yaml
     content_page:
         pages:
             blog_article:
                 path: '/blog/{id}'
+                service: 'blog-articles'
```
giorgiosironi: not obvious that these are paths too (prefixes) rather than e.g. hostnames

vendor-extra/ContentPageBundle/src/Controller/ContentController.php
```diff
@@ -4,12 +4,49 @@
...
+                    'headers' => ['Accept' => 'application/xml'],
+                    'http_errors' => true,
+                ]
+            )
+            ->then(
+                function (ResponseInterface $response) {
+                    $dom = FluentDOM::load((string) $response->getBody());
+                    $dom->registerNamespace('libero', 'http://libero.pub');
+
+                    /** @var string $title */
```
thewilkybarkid: Wonder if we could teach PHPStan XPath. 🤔

PR #19, opened 2018-11-20
Validate Composer

thewilkybarkid:
Like https://github.com/libero/dummy-api/pull/10.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,4 +1,6 @@
 {
+    "name": "libero/browser",
+    "description": "TODO",
     "type": "project",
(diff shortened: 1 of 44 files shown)


composer.json
```diff
@@ -1,4 +1,6 @@
 {
+    "name": "libero/browser",
+    "description": "TODO",
```
thewilkybarkid: Only reason this is WIP.

thewilkybarkid: This will go into the README and GitHub description too.

PR #21, opened 2018-11-21
Add web profiler

thewilkybarkid:


config/bundles.php
```diff
@@ -5,4 +5,6 @@
 return [
     Libero\ContentPageBundle\ContentPageBundle::class => ['all' => true],
     Symfony\Bundle\FrameworkBundle\FrameworkBundle::class => ['all' => true],
+    Symfony\Bundle\TwigBundle\TwigBundle::class => ['all' => true],
```
thewilkybarkid: Not actually needed outside `dev`/`test` yet, but it will be.

config/bundles.php
```diff
@@ -5,4 +5,6 @@
...
+    Symfony\Bundle\WebProfilerBundle\WebProfilerBundle::class => ['dev' => true, 'test' => true],
```
giorgiosironi: why in `test`? Is this just the toolbar, wouldn't expect PHPUnit to use it?

thewilkybarkid: Not needed at the moment, can be used to test things like sent emails (eg https://github.com/elifesciences/journal/blob/238014172fea7fc4c385f782e531f9f96c762708/features/bootstrap/Context.php#L82-L108).

PR #22, opened 2018-11-21
[WIP] Set up Monolog

thewilkybarkid:
Like https://github.com/libero/dummy-api/pull/6.


thewilkybarkid:
Was expecting a 500 error, but there's currently a lack of error handling in the app. This will be enabled by turning on Twig in https://github.com/libero/browser/pull/21 so not going to worry about it.

The messages are logged, which is the important bit here.

PR #23, opened 2018-11-22
Add HEALTHCHECK

giorgiosironi:
Will fail until https://github.com/libero/scripts/pull/2 is integrated.

### DIFF
--- a/.travis/smoke-test.sh
+++ b/.travis/smoke-test.sh
@@ -9,6 +9,6 @@ function finish {
 trap finish EXIT
 
 docker-compose --file docker-compose.yaml up -d web
-.scripts/docker/wait-healthy browser_app_1
+.scripts/docker/wait-healthy.sh browser_app_1
 docker-compose --file docker-compose.yaml exec app bin/console --version
 nc -z localhost 8080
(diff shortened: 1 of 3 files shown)


.travis/smoke-test.sh
```diff
@@ -9,5 +9,6 @@ function finish {
 trap finish EXIT
 
 docker-compose --file docker-compose.yaml up -d web
+.scripts/docker/wait-healthy.sh browser_app_1
 docker-compose --file docker-compose.yaml exec app bin/console --version
 nc -z localhost 8080
```
thewilkybarkid: Technically wants `browser_web_1` too. Maybe we need to be able to wait on multiple containers?

giorgiosironi: By default `nginx` doesn't define an healthcheck:
```
$ docker-compose ps
    Name                   Command                 State              Ports        
-----------------------------------------------------------------------------------
browser_app_1   docker-php-entrypoint php-fpm   Up (healthy)   9000/tcp            
browser_web_1   nginx -g daemon off;            Up             0.0.0.0:8080->80/tcp
```
We could add one in `docker-compose.yaml` and wait for both, though I've never seen such a problem with `nginx` (more on the underlying apps or databases)

thewilkybarkid: Probably not worth the hassle.

PR #24, opened 2018-11-23
Update vendor-extra

thewilkybarkid:
As seen in https://github.com/libero/sample-configuration/pull/10#issuecomment-441251558, `composer update libero/content-page-bundle --with-dependencies` needed to be run to update the lock file.

### DIFF
--- a/.travis/check-vendor-extra-updated.sh
+++ b/.travis/check-vendor-extra-updated.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -e
+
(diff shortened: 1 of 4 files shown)


.travis/check-vendor-extra-updated.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+output=$(docker-compose --file docker-compose.yaml --file docker-compose.test.yaml run app composer update libero/content-page-bundle --with-dependencies --dry-run --no-ansi)
```
thewilkybarkid: Can't find a better way of finding out (`composer outdated` doesn't find it).

PR #25, opened 2018-12-03
Use content header pattern

thewilkybarkid:
Pattern currently manually copied from https://github.com/libero/patterns-core/pull/22.


thewilkybarkid:
Details of overriding Twig templates for experimentation: https://symfony.com/doc/4.1/bundles/override.html#templates

vendor-extra/LiberoPatternsBundle/src/ViewConverter/FrontContentHeaderVisitor.php
```diff
@@ -0,0 +1,61 @@
...
+use DOMNodeList;
+use FluentDOM\DOM\Element;
+use FluentDOM\DOM\Node\ChildNode;
+use Libero\Views\InlineViewConverter;
+use Libero\Views\LangAttributes;
+use Libero\Views\View;
+use Libero\Views\ViewConverterVisitor;
+use function Functional\map;
+
+final class FrontContentHeaderVisitor implements ViewConverterVisitor
```
thewilkybarkid: Need to add tests for these.

vendor-extra/LiberoPatternsBundle/src/ViewConverter/Inline/ItalicVisitor.php
```diff
@@ -0,0 +1,50 @@
...
+    private $inlineConverter;
+
+    public function __construct(InlineViewConverter $inlineConverter)
+    {
+        $this->inlineConverter = $inlineConverter;
+    }
+
+    public function visit(ChildNode $object, View $view, array &$context = []) : View
+    {
+        if (!$object instanceof Element || 'i' !== $object->nodeName || 'http://libero.pub' !== $object->namespaceURI) {
```
thewilkybarkid: Opened https://github.com/ThomasWeinert/FluentDOM/pull/84 to simplify this a little.

Dockerfile
```diff
@@ -37,10 +37,13 @@ COPY .docker/php.ini ${PHP_INI_DIR}/conf.d/00-app.ini
 COPY bin/ bin/
 COPY src/ src/
 COPY public/ public/
+COPY templates/ templates/
 COPY config/ config/
 COPY --from=composer /app/vendor/ vendor/
 COPY vendor-extra/ vendor-extra/
 
+RUN bin/console assets:install && rm -rf var/*
```
giorgiosironi: what's in `var/`, logs/cache?

bin/update-patterns
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
giorgiosironi: needs a description in a comment at the top, what it adds and where it does come from (Github source repository)

bin/update-patterns
```diff
@@ -0,0 +1,33 @@
...
+set -e
+
+function finish {
+    docker rm browser-pattern-library &> /dev/null
```
giorgiosironi: the container may not be existing on exit (if the script errored out sooner than expected), `|| true` may help to make clear this is not the source of the error

vendor-extra/ContentPageBundle/src/Controller/ContentController.php
```diff
@@ -4,24 +4,42 @@
...
+        ClientInterface $client,
+        string $service,
+        Environment $twig,
+        ViewConverter $converter,
+        string $template
+    ) {
         $this->client = $client;
         $this->service = $service;
+        $this->twig = $twig;
+        $this->converter = $converter;
```
giorgiosironi: the number of dependencies here suggests to introduce some collaborator object to hide them, still unclear how to abstract it

vendor-extra/ContentPageBundle/src/Controller/ContentController.php
```diff
@@ -33,18 +51,37 @@ public function __invoke(string $id) : Response
...
+                        throw new UnexpectedValueException('Could not find a front');
                     }
 
-                    return new Response("<html><body>${title}</body></html>");
+                    $context = [
+                        'lang' => $request->getLocale(),
+                        'dir' => 'right-to-left' === Misc::getCharacterOrder($request->getLocale()) ? 'rtl' : 'ltr',
+                    ];
+
+                    $header = $this->converter->convert($front, '@LiberoPatterns/content-header.html.twig', $context);
```
giorgiosironi: sounds like `$this->converter`, `$this->twig` and `$this->template` are tied together but it's early to say how to abstract them into an object until we see more examples

vendor-extra/Views/src/ViewConverterVisitor.php
```diff
@@ -0,0 +1,12 @@
+<?php
```
giorgiosironi: there is a need for documenting what problem this `Visitor` implementation is solving, to make it scalable for other various patterns and content fragments that will come along

Dockerfile
thewilkybarkid: Yep, not things we want to distribute.

thewilkybarkid:
Separated introducing Twig into #27, will then so the same with pulling in the patterns.

vendor-extra/ContentPageBundle/tests/Controller/ContentControllerTest.php
```diff
@@ -38,12 +40,12 @@ public function it_returns_the_title(Request $request, array $twigContext) : voi
                 [],
                 <<<XML
 <?xml version="1.0" encoding="UTF-8"?>
-<item xmlns="http://libero.pub">
-    <front xml:lang="en">
-        <id>id</id>
-        <title>Title</title>
-    </front>
-</item>
+<libero:item xmlns:libero="http://libero.pub">
```
thewilkybarkid: Have to use prefixes as `DOMNode::getNodePath()` doesn't support default namespaces and the element path ends up being `/*/*`. (Refs https://github.com/ThomasWeinert/FluentDOM/issues/61.)

vendor-extra/ContentPageBundle/src/Controller/ContentController.php
thewilkybarkid: `$this->twig` and `$this->template` are definitely tied together.

thewilkybarkid: Leave for now. Might need a little trickery (maybe a template name alias?), but let's see how it plays out first (might be ok to have the same base for all instances, or customisable).

vendor-extra/ContentPageBundle/tests/Controller/ContentControllerTest.php
giorgiosironi: Looks like a big bug for anything real :thinking: 
e.g. this uses the default namespace so it won't work as-is?
```
$ curl http://unstable--dummy-api.libero.pub/scholarly-articles/items/article1/versions/latest
<?xml version="1.0" encoding="UTF-8"?>
<item xmlns="http://libero.pub">
    <front xml:lang="en">
        <id>article1</id>
        <title>Article 1</title>
    </front>
</item>
```

thewilkybarkid:
Was waiting on more tests, but agreed it’s still too big. The templating abstraction can be parked for now as it’s not really required for the milestone, and probably needs more consideration too as it’s getting complex.

vendor-extra/ContentPageBundle/tests/Controller/ContentControllerTest.php
thewilkybarkid: Yep. Unlikely to use in outside tests, but will need to address the issue mentioned if we do.

thewilkybarkid:
Now a lot simpler. Will follow up with a separate PR for the reverted changes.

vendor-extra/ContentPageBundle/src/Controller/ContentController.php
```diff
@@ -49,25 +51,52 @@ function (ResponseInterface $response) use ($request) : Response {
...
                     }
 
                     $context = [
                         'lang' => $request->getLocale(),
-                        'dir' => 'right-to-left' === Misc::getCharacterOrder($request->getLocale()) ? 'rtl' : 'ltr',
+                        'dir' => $this->getDirection($request->getLocale()),
                     ];
 
+                    $contentHeader = [
+                        'template' => '@LiberoPatterns/content-header.html.twig',
```
giorgiosironi: yes, abstracting all the templates that need to be used for each block will be an interesting design to introduce here

vendor-extra/LiberoPatternsBundle/src/Resources/views/italic.html.twig
```diff
@@ -0,0 +1,9 @@
+{% spaceless %}
+
+  <i {%- include '@LiberoPatterns/attributes.html.twig' %}>
```
giorgiosironi: is the `{%-` supposed to strip the `<i ` space or can in principle remain as `<i >` in the empty attributes case? Still valid I think

thewilkybarkid: This comes from the pattern library, but it will strip it. The included template prepends a space if it has a value. Meant to make the template easier to read without resulting in `<i >`.

vendor-extra/LiberoPatternsBundle/src/Resources/views/text.html.twig
```diff
@@ -0,0 +1,17 @@
+{%- if nodes is iterable -%}
```
giorgiosironi: `text` seems a generic name, as it can be confused with the article text for example; perhaps `nodes` or something related could be used as the template name (`all`? `join`?)

vendor-extra/LiberoPatternsBundle/src/Resources/views/content-header.html.twig
```diff
@@ -0,0 +1,40 @@
...
+          {% set lengthClass = 'medium' %}
+        {% elseif length < 80 %}
+          {% set lengthClass = 'long' %}
+        {% elseif length < 120 %}
+          {% set lengthClass = 'x-long' %}
+        {% else %}
+          {% set lengthClass = 'xx-long' %}
+        {% endif %}
+
+        {% set attributes = attributes|default({})|merge({class: attributes.class|default([])|merge(['content-header__title', 'content-header__title--' ~ lengthClass]) }) %}
```
giorgiosironi: the length of this makes it difficult to see the various levels (the second `merge` involving only the `class` attribute); not sure if it will become an helper or if multiple lines can be used to simply reformat it

vendor-extra/LiberoPatternsBundle/src/Resources/views/text.html.twig
thewilkybarkid: Agree that the naming isn't great here (https://github.com/libero/pattern-library/pull/22#discussion_r240106726); one for the pattern library though.

vendor-extra/LiberoPatternsBundle/src/Resources/views/content-header.html.twig
```diff
@@ -0,0 +1,40 @@
...
+
+  {%- block body -%}
+
+    {%- with contentTitle only -%}
+
+      {%- block title -%}
+
+        {% set length = include('@LiberoPatterns/heading.html.twig')|striptags|trim|length %}
+
+        {% if length < 20 %}
```
giorgiosironi: wonder if the magic numbers will be reused elsewhere? early to say

PR #26, opened 2018-12-06
Fix vendor-extra update script

thewilkybarkid:
Currently failing builds as it's also checking other dependencies.

### DIFF
--- a/.travis/check-vendor-extra-updated.sh
+++ b/.travis/check-vendor-extra-updated.sh
@@ -1,7 +1,18 @@
 #!/bin/bash
 set -e
 
-output=$(docker-compose --file docker-compose.yaml --file docker-compose.test.yaml run app composer update libero/content-page-bundle --with-dependencies --dry-run --no-ansi)
+function finish {
+    docker-compose --file docker-compose.yaml --file docker-compose.test.yaml down --volumes
+}
+
+trap finish EXIT
+
+packages=("libero/content-page-bundle" "libero/patterns-bundle" "libero/views")
+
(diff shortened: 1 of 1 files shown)


.travis/check-vendor-extra-updated.sh
```diff
@@ -1,7 +1,18 @@
...
 set -e
 
-output=$(docker-compose --file docker-compose.yaml --file docker-compose.test.yaml run app composer update libero/content-page-bundle --with-dependencies --dry-run --no-ansi)
+function finish {
+    docker-compose --file docker-compose.yaml --file docker-compose.test.yaml down --volumes
+}
+
+trap finish EXIT
+
+packages=("libero/content-page-bundle" "libero/patterns-bundle" "libero/views")
```
thewilkybarkid: Should be able to work this out.

.travis/check-vendor-extra-updated.sh
```diff
@@ -1,7 +1,20 @@
...
+
+while IFS=$'\n' read -r line; do packages+=("$line"); done < <(< composer.lock jq --raw-output '.packages[] | select(.dist.url | startswith("./vendor-extra/")) .name')
+
+output=$(docker-compose --file docker-compose.yaml --file docker-compose.test.yaml run app composer update --dry-run --no-ansi)
 echo "${output}"
 
-[[ ${output} =~ "Nothing to install or update" ]]
+for package in "${packages[@]}";
+do
+    [[ ${output} != *"Updating ${package}"* ]]
```
giorgiosironi: The output of this does not say much, could try adding `set -x` at the start of the script? Or making this the condition an explicit `if` that prints an error message before `exit $someNumber`

PR #27, opened 2018-12-12
Use Twig

thewilkybarkid:
Extracted from #25.

### DIFF
--- a/tests/WebTest.php
+++ b/tests/WebTest.php
@@ -6,6 +6,7 @@ namespace tests\Libero\Browser;
 
 use GuzzleHttp\Psr7\Request;
 use GuzzleHttp\Psr7\Response;
+use function trim;
 
 final class WebTest extends WebTestCase
 {
@@ -33,7 +34,6 @@ final class WebTest extends WebTestCase
         <id>{$id}</id>
         <title>Scholarly article {$id}</title>
     </front>
-
 </item>
 XML
             )
@@ -44,7 +44,7 @@ XML
 
         $this->assertSame(200, $response->getStatusCode());
         $this->assertSame('text/html; charset=UTF-8', $response->headers->get('Content-Type'));
-        $this->assertSame("Scholarly article {$id}", $crawler->text());
+        $this->assertSame("Scholarly article {$id}", trim($crawler->filter('body')->text()));
     }
 
     /**
@@ -81,7 +81,7 @@ XML
 
         $this->assertSame(200, $response->getStatusCode());
         $this->assertSame('text/html; charset=UTF-8', $response->headers->get('Content-Type'));
-        $this->assertSame("Blog article {$id}", $crawler->text());
+        $this->assertSame("Blog article {$id}", trim($crawler->filter('body')->text()));
     }
 
     public function idProvider() : iterable
--- a/vendor-extra/ContentPageBundle/src/Controller/ContentController.php
+++ b/vendor-extra/ContentPageBundle/src/Controller/ContentController.php
@@ -7,21 +7,33 @@ namespace Libero\ContentPageBundle\Controller;
 use FluentDOM;
 use GuzzleHttp\ClientInterface;
 use Psr\Http\Message\ResponseInterface;
+use Punic\Misc;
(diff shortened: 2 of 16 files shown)


vendor-extra/ContentPageBundle/src/Controller/ContentController.php
```diff
@@ -7,21 +7,33 @@
...
+    private $template;
+    private $twig;
 
-    public function __construct(ClientInterface $client, string $service)
-    {
+    public function __construct(
+        ClientInterface $client,
+        string $service,
+        Environment $twig,
+        string $template
```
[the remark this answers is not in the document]
thewilkybarkid: Can be done later though.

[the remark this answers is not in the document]
thewilkybarkid: Relates to https://github.com/libero/browser/pull/25#discussion_r240235839 and https://github.com/libero/browser/pull/25#discussion_r240237354.

vendor-extra/ContentPageBundle/src/Controller/ContentController.php
```diff
@@ -44,7 +56,22 @@ function (ResponseInterface $response) {
                         throw new UnexpectedValueException('Could not find a title');
                     }
 
-                    return new Response("<html><body>${title}</body></html>");
+                    $context = [
+                        'lang' => $request->getLocale(),
+                        'dir' => 'right-to-left' === Misc::getCharacterOrder($request->getLocale()) ? 'rtl' : 'ltr',
```
giorgiosironi: `Misc` class :face_with_head_bandage:

vendor-extra/ContentPageBundle/tests/Controller/ContentControllerTest.php
```diff
@@ -47,13 +49,24 @@ public function it_returns_the_title(string $id) : void
...
+                [
+                    'template.html.twig',
+                    [
+                        'lang' => 'en',
+                        'dir' => 'ltr',
+                        'title' => "Article {$id}",
+                    ],
+                ]
+            ),
+            $response->getContent()
```
giorgiosironi: `text/html; charset=UTF-8` but it's a JSON string? :thinking:

giorgiosironi: ah, `TwigTestCase` doesn't really render the template because we are in the context of a bundle; confusing, could use a single HTML element in which the JSON is dumped (and would leave space to dump other things)

tests/WebTest.php
```diff
@@ -81,7 +81,7 @@ public function it_shows_blog_articles(string $id) : void
 
         $this->assertSame(200, $response->getStatusCode());
         $this->assertSame('text/html; charset=UTF-8', $response->headers->get('Content-Type'));
-        $this->assertSame("Blog article {$id}", $crawler->text());
+        $this->assertSame("Blog article {$id}", trim($crawler->filter('body')->text()));
```
giorgiosironi: good for now, will have to look into a certain element (`h1`?) when that is used

thewilkybarkid: Will become https://github.com/libero/browser/blob/1bc82c5dd9e95a9f3eaf49ff44f47b6e85aa3e8f/tests/WebTest.php#L83 in #25.

vendor-extra/ContentPageBundle/tests/Controller/ContentControllerTest.php
thewilkybarkid: Makes it harder to compare (string matching could be brittle). Agree it looks a bit confusing on first glance.

giorgiosironi: String matching is brittle, but it's possible to match the element and extract the inner text to parse it with `json_decode`; it doesn't look like this is a test that needs to be replicated many times, and even so the access to the right element can be extracted into a private method provided by `TwigTestCase` itself to keep all the knowledge about this in a single place

thewilkybarkid:
Should be good to go now @giorgiosironi.

PR #28, opened 2018-12-13
Add styling

thewilkybarkid:
Extracted from #25. Should be merged after https://github.com/libero/pattern-library/pull/26.


thewilkybarkid:
Should be good to go.

Dockerfile
```diff
@@ -78,6 +81,9 @@ COPY composer.json \
     ./
 COPY --from=composer-dev /app/vendor/ vendor/
 
+RUN bin/console assets:install && \
```
giorgiosironi: what was the reason for installing twice, the files from the `prod` target can't get here?

bin/update-patterns
```diff
@@ -0,0 +1,33 @@
...
+trap finish EXIT
+
+cd "$(dirname "$0")/.."
+
+mkdir -p var/pattern-library && cd "$_"
+
+if ! [[ -d .git ]]; then
+    git clone git@github.com:libero/pattern-library.git .
+else
+    git reset --hard origin/master
```
giorgiosironi: is there a potential for this script to work on a branch of `pattern-library`?

thewilkybarkid: Not particularly, only running the build and export commands which isn’t much duplication. Plus this should be replaced by using ready-built artifacts anyway.

Dockerfile
thewilkybarkid: Might be different per environment, depending on what’s enabled. Thinking about it, should do it on `dev` as well (and `test` is redundant until there’s a test that goes through a real browser).

giorgiosironi:
What should we see at http://unstable.libero.pub/articles/article1?
I don't get an `<h1>` there, seems similar to the previous version.
The versions in `.env` seem correct:
```
# TODO: put this to latest as a default, when `latest` is available
REVISION_BROWSER=488fccea376b09149de6874edfc0ac7094d06b14
# TODO: put this to latest as a default, when `latest` is available
REVISION_DUMMY_API=f0fdadac9163b26232f7b24cc77656ce1417c20d
```
and the containers too:
```
ubuntu@ip-172-31-70-78:/tmp/sample-configuration$ docker ps | grep browser
1af40090fdf2        liberoadmin/browser:488fccea376b09149de6874edfc0ac7094d06b14     "docker-php-entrypoi…"   3 hours ago         Up 3 hours (healthy)   9000/tcp             sample-configuration_browser_fpm_1
```

giorgiosironi:
Ah right, https://github.com/libero/browser/pull/25 is open

PR #29, opened 2018-12-19
Abstract creation of the patterns

thewilkybarkid:
Initially contains the changes reverted in https://github.com/libero/browser/pull/25.


thewilkybarkid:
Now includes #30 to try combining them.

vendor-extra/JatsContentBundle/src/ViewConverter/FrontContentHeaderVisitor.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        return '@LiberoPatterns/content-header.html.twig';
+    }
+
+    protected function expectedElement() : string
+    {
+        return '{http://jats.nlm.nih.gov}front';
+    }
+
+    protected function unexpectedArguments() : array
```
giorgiosironi: this seems to be a consolidated way to avoid accidentally overwriting an existing view argument, but I'm not sure what decides the priority of the Visitors

vendor-extra/ViewsBundle/src/Views/SimplifiedVisitor.php
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        $expectedElement = $this->expectedElement();
+        $actualElement = sprintf('{%s}%s', $object->namespaceURI, $object->localName);
+
+        if (is_string($expectedElement) && $expectedElement !== $actualElement) {
+            return $view;
+        }
+
+        foreach ($this->unexpectedArguments() as $argument) {
```
giorgiosironi: the unexpected arguments also seem to be in `and`: as soon as one appears the Visitor does nothing. They could be in `or` as well, but we don't have an example where we have two?

vendor-extra/JatsContentBundle/src/ViewConverter/ArticleTitleHeadingVisitor.php
```diff
@@ -0,0 +1,35 @@
...
+use Libero\ViewsBundle\Views\View;
+use Libero\ViewsBundle\Views\ViewConverterVisitor;
+
+final class ArticleTitleHeadingVisitor implements ViewConverterVisitor
+{
+    use SimplifiedVisitor;
+
+    protected function doVisit(Element $object, View $view, array &$context = []) : View
+    {
+        return $view->withArgument('text', (string) $object);
```
giorgiosironi: perhaps by decorating the `View` object, but as the population of Visitor grows there's the need for a standard way to log which `Visitor` has done what , to operate this correctly and understand what is clashing (if anything)

composer.json
```diff
@@ -13,7 +13,8 @@
             "tests\\Libero\\Browser\\": "tests/",
             "tests\\Libero\\ContentPageBundle\\": "vendor-extra/ContentPageBundle/tests/",
             "tests\\Libero\\JatsContentBundle\\": "vendor-extra/JatsContentBundle/tests/",
-            "tests\\Libero\\LiberoContentBundle\\": "vendor-extra/LiberoContentBundle/tests/"
+            "tests\\Libero\\LiberoContentBundle\\": "vendor-extra/LiberoContentBundle/tests/",
+            "tests\\Libero\\ViewsBundle\\": "vendor-extra/ViewsBundle/tests/"
```
giorgiosironi: what is the direction of dependencies? Format-specific bundles depend on `ViewsBundle` where the interfaces are?

That module really needs a README to document common cases for adding/modifying Visitors.

vendor-extra/JatsContentBundle/src/ViewConverter/FrontContentHeaderVisitor.php
thewilkybarkid: The service tag https://github.com/libero/browser/blob/c7a401521a4dca7fe15bc9b0fae691bebe84e860/vendor-extra/ViewsBundle/src/Resources/config/services.xml#L19 (one for documentation)

vendor-extra/ViewsBundle/src/Views/SimplifiedVisitor.php
thewilkybarkid: Yeah, not entirely sure about this yet.

vendor-extra/JatsContentBundle/src/ViewConverter/ArticleTitleHeadingVisitor.php
thewilkybarkid: Had thought the same, some mechanism to record this in the Symfony Profiler.

thewilkybarkid:
Merging to unblock other tickets. Might need refactoring, and definitely needs documentation.

PR #30, opened 2019-01-11
Support JATS articles

thewilkybarkid:
Complementary to #29, and based on the possible idea of not requiring content items to be a `{http://libero.pub}item`.

This changes the scholarly article example to be `{http://jats.nlm.nih.gov}article`.


vendor-extra/ContentPageBundle/src/Controller/ContentController.php
```diff
@@ -49,56 +49,16 @@ public function __invoke(Request $request, string $id) : Response
...
                         $this->twig->render(
                             $this->template,
-                            array_merge(
-                                $context,
-                                [
-                                    'title' => $title,
-                                    'content' => [$contentHeader],
-                                ]
-                            )
+                            $this->contentHandler->handle($dom->documentElement, $context)
```
thewilkybarkid: Other option is not have separate controllers for each type.

vendor-extra/ContentPageBundle/src/Handler/JatsContentHandler.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Libero\ContentPageBundle\Handler;
```
thewilkybarkid: Should go into a separate JATS bundle.

vendor-extra/ContentPageBundle/src/functions.php
```diff
@@ -0,0 +1,15 @@
...
+declare(strict_types=1);
+
+namespace Libero\ContentPageBundle;
+
+use Punic\Misc;
+
+const DIRECTION_LTR = 'ltr';
+const DIRECTION_RTL = 'rtl';
+
+function text_direction(string $locale) : string
```
thewilkybarkid: Should be extracted from this PR. Not entirely sure of the best place for it...

.docker/api/data/scholarly-articles/article1/1.xml
```diff
@@ -1,12 +1,27 @@
...
 
-    <front xml:lang="en">
+    <meta>
 
         <id>article1</id>
-        <title>Article 1</title>
 
-    </front>
+    </meta>
+
```
giorgiosironi: the wrapper XML format itself makes sense

vendor-extra/JatsContentBundle/src/Handler/JatsContentHandler.php
```diff
@@ -0,0 +1,98 @@
...
+use FluentDOM\DOM\Attribute;
+use FluentDOM\DOM\Document;
+use FluentDOM\DOM\Element;
+use FluentDOM\DOM\Xpath;
+use Libero\ContentPageBundle\Handler\ContentHandler;
+use UnexpectedValueException;
+use function array_merge;
+use function Libero\ContentPageBundle\text_direction;
+
+final class JatsContentHandler implements ContentHandler
```
giorgiosironi: the name `JatsContentHandler` (and perhaps `JatsContentBundle`) is misleading if, as far as JATS variations go, this can support eLife-like JATS or Libero-constrained JATS rather than the totality of JATS content. This makes a conversion step still necessary from non-Libero-constrained JATS to Libero-constrained JATS, _unless_ this frontend and all the various services make efforts to support various options.

giorgiosironi: On the ingestion validation side of things, I guess a Schema(-tron?) could be written to validate it's Libero-constrained JATS, don't know how difficult it is to do

thewilkybarkid: Indeed, not yet clear _what_ JATS will be supported yet. I suspect it's going to be somewhere between DAR and JATS4R. There might be a certain amount of configuration too, so you could tweak selectors.

thewilkybarkid: It's not. You probably could use XSLT to do the same (presumably producing JSON which is a valid input for the patterns), but I suspect there's things not possible there (eg [having to replace placeholders in code afterwards](https://github.com/elifesciences/elife-ingest-xsl/blob/bd0e114f26af8a9fb506ac3ee6bb41d28d54d3fa/lib/xsl/jats-to-html.xsl#L1467-L1468)).

vendor-extra/ContentPageBundle/src/Controller/ContentController.php
giorgiosironi: This has been updated to use a single controller, it seems

vendor-extra/ContentPageBundle/src/DependencyInjection/Compiler/ContentHandlerPass.php
```diff
@@ -0,0 +1,40 @@
...
+namespace Libero\ContentPageBundle\DependencyInjection\Compiler;
+
+use LogicException;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+use function array_keys;
+
+final class ContentHandlerPass implements CompilerPassInterface
```
giorgiosironi: these `Passes` could do with a docblock describing how they map to the YAML schema, but maybe this is already described elsewhere

vendor-extra/ContentPageBundle/src/DependencyInjection/ContentPageExtension.php
```diff
@@ -47,6 +47,7 @@ private function addPage(string $name, array $config, ContainerBuilder $containe
         $definition->setArgument(2, new Reference('twig'));
         $definition->setArgument(3, $config['page_template']);
         $definition->addTag('controller.service_arguments');
+        $definition->addTag('libero.content_page.controller', ['handler' => $config['handler']]);
```
giorgiosironi: `handler` is overloaded, we'll find a metaphor for the naming I suppose, given publishing wealth of concepts and strange words

PR #31, opened 2019-02-01
Standardise use of CallbackViewConverter in tests

thewilkybarkid:


vendor-extra/ContentPageBundle/tests/ViewConvertingTestCase.php
```diff
@@ -8,14 +8,33 @@
...
+                if ($includeTemplate) {
+                    $arguments['template'] = $template;
+                }
+
+                return new View($template, $arguments);
+            }
+        );
+    }
+
+    final protected function createFailingConverter() : ViewConverter
```
giorgiosironi: good extraction of duplicated code

vendor-extra/ContentPageBundle/tests/ViewConvertingTestCase.php
```diff
@@ -8,14 +8,33 @@
 use Libero\ViewsBundle\Views\CallbackViewConverter;
 use Libero\ViewsBundle\Views\View;
 use Libero\ViewsBundle\Views\ViewConverter;
+use LogicException;
 
 trait ViewConvertingTestCase
 {
-    final protected function createConverter() : ViewConverter
+    final protected function createConverter(bool $includeTemplate = false) : ViewConverter
```
giorgiosironi: boolean argument (especially without keyword-based arguments like in Python) are not expressive, consider `$this->createConverter(self::INCLUDE_TEMPLATE)` which can still be a boolean but expressed as a constant. Very low cost for readability

vendor-extra/LiberoContentBundle/tests/ViewConverter/FrontContentHeaderVisitorTest.php
```diff
@@ -153,7 +130,7 @@ function (Element $object, ?string $template, array $context) : View {
         $this->assertEquals(
             [
                 'contentTitle' => [
-                    'object' => $element->childNodes->item(0),
+                    'element' => '/libero:front/libero:title',
```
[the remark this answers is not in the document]
giorgiosironi: To clarify, it's a bit confusing that this format (`element` rather than `object`) it's part of the test cases through `ViewConvertingTestCase`, as reading it in the expected field I was expecting it to have changed somewhere in the production code.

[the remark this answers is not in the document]
giorgiosironi: I would rename `createConverter` to something more specific like `createDumpingConverter` or similar, since it introduces the custom code that we rely upon in these assertions

vendor-extra/ContentPageBundle/tests/ViewConvertingTestCase.php
thewilkybarkid: Agreed. Since traits can't have constants, just going to remove the option.

PR #32, opened 2019-02-01
Inline templating

thewilkybarkid:
Includes #31.


vendor-extra/ViewsBundle/src/Views/InlineViewConverterRegistry.php
```diff
@@ -0,0 +1,46 @@
...
+    public function convert(NonDocumentTypeChildNode $object, array $context = []) : View
+    {
+        $view = new View(null, []);
+
+        foreach ($this->visitors as $visitor) {
+            $view = $visitor->visit($object, $view, $context);
+        }
+
+        if (!$view->getTemplate()) {
+            return new View('@LiberoPatterns/text.html.twig', ['nodes' => (string) $object]);
```
thewilkybarkid: This couples the `ViewsBundle` to the `LiberoPatternsBundle`, should probably throw an exception instead and have a late running listener somewhere else. Can leave for now though.

vendor-extra/ViewsBundle/src/Views/SimplifiedVisitor.php
```diff
@@ -29,12 +32,28 @@
...
 
     abstract protected function doVisit(Element $object, View $view, array &$context = []) : View;
 
-    abstract protected function expectedTemplate() : string;
+    protected function expectedTemplate() : ?string
+    {
+        return null;
+    }
+
+    protected function possibleTemplate() : ?string
```
thewilkybarkid: Should split into 2 traits.

vendor-extra/ViewsBundle/src/Views/SimplifiedInlineVisitor.php
```diff
@@ -0,0 +1,51 @@
...
+
+declare(strict_types=1);
+
+namespace Libero\ViewsBundle\Views;
+
+use FluentDOM\DOM\Element;
+use function is_string;
+use function sprintf;
+
+trait SimplifiedInlineVisitor
```
thewilkybarkid: Need a better name, would also apply to thinks like paragraphs (another thing where you don't know the template up front) which aren't inline.

vendor-extra/ViewsBundle/src/Views/InlineViewConverter.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Libero\ViewsBundle\Views;
+
+use FluentDOM\DOM\Node\NonDocumentTypeChildNode;
+
+interface InlineViewConverter
```
thewilkybarkid: Wondering whether this should be separate at all. There's three types of things to convert:

1. Chunk of data into a known pattern (eg front matter to content header)
2. List of mixed content into unknown patterns (eg mixed text and inline elements into a variety of patterns)
3. Chunks of data into unknown patterns (eg list of paragraphs, figures etc into a variety of patterns)

Currently 1 and 2 have dedicated interfaces, rather than introduce a third we can merge them all into 1:

```php
interface ViewConverter
{
    public function convert(NonDocumentTypeChildNode $object, ?string $template = null, array $context = []) : View;
}
```

Then the implementation filters out non-elements, like https://github.com/libero/browser/blob/12e7c1d0bb224e386236a958fb6932e10d7646ba/vendor-extra/ViewsBundle/src/Views/InlineViewConverterRegistry.php#L22-L24 (but also checking the template).

thewilkybarkid: At some point we should find optimisations for the visitors (eg split them by expected template, element).

vendor-extra/JatsContentBundle/tests/ViewConverter/ArticleTitleHeadingVisitorTest.php
```diff
@@ -85,17 +88,32 @@ public function it_does_nothing_if_there_is_already_text_set() : void
...
 
         $newContext = ['foo' => 'bar'];
         $view = $visitor->visit($element, new View('@LiberoPatterns/heading.html.twig'), $newContext);
 
         $this->assertSame('@LiberoPatterns/heading.html.twig', $view->getTemplate());
-        $this->assertEquals(['text' => 'foo baz'], $view->getArguments());
+        $this->assertEquals(
+            [
+                'text' => [
+                    new View(null, ['object' => '/jats:article-title/text()[1]', 'context' => ['foo' => 'bar']]),
```
giorgiosironi: too much `foo bar` doesn't explain where things go; I am mapping this to 
```
foo <jats:italic>bar</jats:italic>
```
which means... the text before italic becomes the key of a `context` array value, which is the content of the italic tag? :thinking:

thewilkybarkid: Heh, will change the context names.

vendor-extra/LiberoContentBundle/tests/ViewConverter/Inline/ItalicVisitorTest.php
```diff
@@ -0,0 +1,115 @@
...
+
+        $this->assertSame('template', $view->getTemplate());
+        $this->assertEmpty($view->getArguments());
+        $this->assertEmpty($newContext);
+    }
+
+    /**
+     * @test
+     */
+    public function it_does_nothing_if_there_is_already_text_set() : void
```
giorgiosironi: I'm not sure how the `it_does_nothing_if_there_is_already_text_set` tests category works; it means that the content gets ignored? I suppose because there is an equivalent `text` instead, so should `'bar'` here be more similar to what it is overriding?

vendor-extra/JatsContentBundle/tests/ViewConverter/ArticleTitleHeadingVisitorTest.php
giorgiosironi: Incidentally, if the various `foo` are not meant to be the same thing, mutation testing may find it as a test not failing when variables are mixed up (but it's a brute force approach, and other tests may fail anyway)

vendor-extra/LiberoContentBundle/tests/ViewConverter/Inline/ItalicVisitorTest.php
thewilkybarkid: Would making it `foo` be clearer? So you have a visitor with a high priority that has already processed the text (presumably different in some way) and haven't disabled this one (so the check stops it overwriting).

giorgiosironi: Is the assumption that the text is only processed by one visitor too strong? You could have the title into a `<title>` and `<h1>` tags (silly example but I'm sure there are similar cases).

thewilkybarkid: There's two different things there. Processing the text for this case (yes, I'd expect only one visitor to do hit, hence these bailing out), and for the title appearing in multiple cases on the page (better example is the author list, which is in at least 3 places for eLife: content header, author information section and metadata).

vendor-extra/ViewsBundle/src/Views/InlineViewConverter.php
thewilkybarkid: I've pushed this change in bca13a4, still open for discussion.

vendor-extra/JatsContentBundle/tests/ViewConverter/FrontContentHeaderVisitorTest.php
```diff
@@ -150,20 +150,20 @@ public function it_sets_the_text_argument() : void
         /** @var Element $element */
         $element = $xml->documentElement;
 
-        $newContext = ['foo' => 'bar'];
+        $newContext = ['bar' => 'baz'];
```
nlisgo: What was the reason for this change?

PR #33, opened 2019-02-05
Add categories to content headers

nlisgo:


vendor-extra/JatsContentBundle/src/ViewConverter/ArticleCategoriesVisitor.php
```diff
@@ -0,0 +1,44 @@
...
+use Libero\ViewsBundle\Views\ViewConverterVisitor;
+
+final class ArticleCategoriesVisitor implements ViewConverterVisitor
+{
+    use SimplifiedVisitor;
+
+    protected function doVisit(Element $object, View $view, array &$context = []) : View
+    {
+        // @fixme - just experimental.
+        return $view->withArgument('items',
```
nlisgo: This is not getting picked up. What am I missing?

vendor-extra/JatsContentBundle/src/ViewConverter/ArticleCategoriesVisitor.php
```diff
@@ -0,0 +1,44 @@
...
+                        'href' => '#',
+                    ],
+                    'text' => 'Cell Biology',
+                ],
+            ]);
+    }
+
+    protected function expectedTemplate() : string
+    {
+        return 'tag-list.html.twig';
```
thewilkybarkid:
```suggestion
        return '@LiberoPatterns/tag-list.html.twig';
```

vendor-extra/JatsContentBundle/src/ViewConverter/ArticleCategoriesVisitor.php
thewilkybarkid: Probably https://github.com/libero/browser/pull/33#discussion_r253848369.

vendor-extra/JatsContentBundle/src/ViewConverter/FrontContentHeaderVisitor.php
```diff
@@ -31,13 +31,19 @@ protected function doVisit(Element $object, View $view, array &$context = []) :
 
         $title = $xpath->firstOf('jats:article-meta/jats:title-group/jats:article-title', $object);
 
+        $categories = $xpath->firstOf('jats:article-meta/jats:article-categories/jats:subj-group[@subj-group-type = "heading"]/jats:subject/ancestor::jats:article-categories', $object);
```
thewilkybarkid: Raises the interesting question about how granular each of these visitors are. Thinking that they should each done one thing, so they can easily be turned off/replaced individually (maybe you want to do something else completely with the categories list). So this one should be renamed `FrontArticleTitleContentHeaderVisitor`, and this part moved to `FrontSubjectGroupContentHeaderVisitor` (no need for the backtracking).

vendor-extra/JatsContentBundle/src/ViewConverter/ArticleCategoriesVisitor.php
```diff
@@ -0,0 +1,44 @@
...
+    use SimplifiedVisitor;
+
+    protected function doVisit(Element $object, View $view, array &$context = []) : View
+    {
+        // @fixme - just experimental.
+        return $view->withArgument('items',
+            [
+                'content' => [
+                    'attributes' => [
+                        'href' => '#',
```
thewilkybarkid: Ignore links for now.

vendor-extra/JatsContentBundle/src/ViewConverter/FrontContentHeaderVisitor.php
nlisgo: I will rename `FrontArticleTitleContentHeaderVisitor` as a followup to this?

vendor-extra/JatsContentBundle/src/ViewConverter/FrontSubjectGroupContentHeaderVisitor.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    protected function doVisit(Element $object, View $view, array &$context = []) : View
+    {
+        /** @var Document $document */
+        $document = $object->ownerDocument;
+        $xpath = $document->xpath();
+        $xpath->registerNamespace('jats', 'http://jats.nlm.nih.gov');
+
+        // @todo - add support for other ways of expressing a subject group in JATS.
```
thewilkybarkid: This all comes down to the 'which JATS' question. For now we'll keep just this way (since you can always write code  yourself to add others).

vendor-extra/JatsContentBundle/src/ViewConverter/FrontSubjectGroupContentHeaderVisitor.php
```diff
@@ -0,0 +1,74 @@
...
+        $groups = $xpath->evaluate(implode('/', [
+            'jats:article-meta/jats:article-categories',
+            'jats:subj-group[@subj-group-type = "heading"]/jats:subject',
+        ]), $object);
+
+        if (!$groups instanceof DOMNodeList || $groups->count() === 0) {
```
thewilkybarkid: I'd prefer https://github.com/libero/browser/blob/fddd0f99cbf4aeb3447e1b73b04a441e46991c49/vendor-extra/JatsContentBundle/src/ViewConverter/FrontItemTagsVisitor.php#L35-L38 syntax.

(If https://github.com/phpstan/phpstan/pull/1752 is accepted PHPStan will understand the return type, but for now we can just annotated it. Also prefer the `count` function.)

vendor-extra/JatsContentBundle/src/ViewConverter/FrontSubjectGroupContentHeaderVisitor.php
```diff
@@ -0,0 +1,74 @@
...
+
+    protected function doVisit(Element $object, View $view, array &$context = []) : View
+    {
+        /** @var Document $document */
+        $document = $object->ownerDocument;
+        $xpath = $document->xpath();
+        $xpath->registerNamespace('jats', 'http://jats.nlm.nih.gov');
+
+        // @todo - add support for other ways of expressing a subject group in JATS.
+        $groups = $xpath->evaluate(implode('/', [
```
thewilkybarkid: `$subjects`?

vendor-extra/JatsContentBundle/src/ViewConverter/FrontSubjectGroupContentHeaderVisitor.php
```diff
@@ -0,0 +1,74 @@
...
+
+        if (!$groups instanceof DOMNodeList || $groups->count() === 0) {
+            return $view;
+        }
+
+        $items = [];
+        foreach ($groups as $group) {
+            $items[] = [
+                'content' => [
+                    'text' => $this->convertChildren($group, $context),
```
thewilkybarkid: While we are just having text for now, we should use the `link` pattern which would allow for someone to add a `href`, like https://github.com/libero/browser/blob/fddd0f99cbf4aeb3447e1b73b04a441e46991c49/vendor-extra/JatsContentBundle/src/ViewConverter/KeywordGroupTagListVisitor.php#L61-L74

vendor-extra/JatsContentBundle/tests/ViewConverter/FrontSubjectGroupContentHeaderVisitorTest.php
```diff
@@ -0,0 +1,98 @@
...
+            $element,
+            new View('@LiberoPatterns/content-header.html.twig'),
+            $newContext
+        );
+
+        $this->assertSame('@LiberoPatterns/content-header.html.twig', $view->getTemplate());
+        $this->assertEmpty($view->getArguments());
+        $this->assertEmpty($newContext);
+    }
+}
```
thewilkybarkid: We'll need tests here to cover a positive result, like https://github.com/libero/browser/blob/fddd0f99cbf4aeb3447e1b73b04a441e46991c49/vendor-extra/JatsContentBundle/tests/ViewConverter/FrontItemTagsVisitorTest.php#L131-L181

vendor-extra/JatsContentBundle/src/ViewConverter/FrontContentHeaderVisitor.php
thewilkybarkid: Yeah, or can get it in now.

thewilkybarkid:
I've copied in some functionality from #35, so that'll need to be merged first.

vendor-extra/JatsContentBundle/src/ViewConverter/FrontSubjectGroupContentHeaderVisitor.php
```diff
@@ -0,0 +1,80 @@
...
+
+        if (0 === count($subjects)) {
+            return $view;
+        }
+
+        return $view->withArgument(
+            'categories',
+            [
+                'attributes' => [
+                    'aria-label' => $this->translate('libero.patterns.content_header.categories.label', $context),
```
thewilkybarkid: This should actually be a separate visitor that runs late, so it's easy to override (you probably want something more specific).

thewilkybarkid: Will do that as a follow up.

PR #34, opened 2019-02-05
Add paragraphs of text

thewilkybarkid:


.docker/api/data/scholarly-articles/article1/1.xml
```diff
@@ -15,13 +15,91 @@
             <jats:article-meta>
 
                 <jats:title-group>
-                    <jats:article-title>Article 1</jats:article-title>
+
+                    <jats:article-title>
+                        <jats:italic>Homo naledi</jats:italic>, a new species of the genus
+                        <jats:italic>Homo</jats:italic> from the Dinaledi Chamber, South Africa
+                    </jats:article-title>
```
nlisgo: This change is unrelated to paragraphs but I welcome it.

vendor-extra/JatsContentBundle/src/Handler/JatsContentHandler.php
```diff
@@ -46,11 +48,22 @@ public function handle(Element $documentElement, array $context) : array
 
         $contentHeader = $this->converter->convert($front, '@LiberoPatterns/content-header.html.twig', $context);
 
+        $content = [$contentHeader];
```
nlisgo: Not blocking: since `$contentHeader` is not being used again might be cleaner to?:

```
$content = [
    $this->converter->convert($front, '@LiberoPatterns/content-header.html.twig', $context),
];
```

vendor-extra/LiberoPatternsBundle/src/Resources/views/content-header.html.twig
```diff
@@ -4,6 +4,16 @@
 
   {%- block body -%}
 
+    {%- with categories|default({}) only -%}
```
nlisgo: Not blocking: Have we considered automatically introducing the latest patterns in line with them being deployed in pattern library to reduce the delta for these PR's?

vendor-extra/JatsContentBundle/src/Handler/JatsContentHandler.php
thewilkybarkid: It's used in https://github.com/libero/browser/blob/b6ebf895424c7b62a0587188f5b26343bb3dc998/vendor-extra/JatsContentBundle/src/Handler/JatsContentHandler.php#L65

vendor-extra/LiberoPatternsBundle/src/Resources/views/content-header.html.twig
thewilkybarkid: We will do, not sure when's the best time. The issue at the minute is that the patterns were merged in a different order to how they're being implemented. Given that things are unblocked hopefully we can keep the order reasonably consistent which will mitigate it.

vendor-extra/JatsContentBundle/src/Handler/JatsContentHandler.php
nlisgo: Ignore me!

PR #35, opened 2019-02-05
Add item tags

thewilkybarkid:


vendor-extra/ContentPageBundle/src/functions.php
```diff
@@ -19,3 +23,16 @@ function text_direction(string $locale) : string
...
+{
+    return sprintf($text, ...array_map('Libero\ContentPageBundle\normalize_translation_key_part', $arguments));
+}
+
+/**
+ * @internal
+ */
+function normalize_translation_key_part($part) : string
+{
+    return strtolower(str_replace('-', '_', (string) $part));
```
thewilkybarkid: Rather naïve; assumes the character range is `[A-z-_]`.

vendor-extra/JatsContentBundle/src/ViewConverter/KeywordGroupTagListVisitor.php
```diff
@@ -0,0 +1,101 @@
...
+                'title',
+                $this->converter->convert($title, '@LiberoPatterns/heading.html.twig', $context)->getArguments()
+            );
+        } else {
+            $view = $view->withArgument(
+                'title',
+                [
+                    'text' => $this->translator->trans(
+                        translation_key(
+                            'libero.jats.keyword_group.title.%s',
```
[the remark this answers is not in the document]
thewilkybarkid: Other option is to have a mapping of known `@kwd-group-type` to translation keys, then only display these groups if there's a `<jats:title>` or a known `@kwd-group-type`.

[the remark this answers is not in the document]
thewilkybarkid: I removed the dynamic keys and added a fallback title in 0dcfc396b7f69e36579d432a71adbaa5f9c6fc81.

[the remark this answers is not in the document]
thewilkybarkid: Decided to skip in 06e8cde2af42b9881c078c5da225ac2ba203da15, but make the map of keys to titles configurable.

vendor-extra/JatsContentBundle/src/Handler/JatsContentHandler.php
```diff
@@ -46,11 +47,21 @@ public function handle(Element $documentElement, array $context) : array
 
         $contentHeader = $this->converter->convert($front, '@LiberoPatterns/content-header.html.twig', $context);
 
+        $content = [$contentHeader];
+
+        $tags = $this->converter->convert($front, '@LiberoPatterns/item-tags.html.twig', $context);
+        if ($tags->hasArgument('groups')) {
```
thewilkybarkid: Awkward to add into `JatsContentHandlerTest`, as the dumping converted won't set the real arguments...

vendor-extra/JatsContentBundle/src/ViewConverter/HeadingVisitor.php
```diff
@@ -31,11 +31,6 @@ protected function expectedTemplate() : string
         return '@LiberoPatterns/heading.html.twig';
     }
 
-    protected function expectedElement() : string
-    {
-        return '{http://jats.nlm.nih.gov}article-title';
```
thewilkybarkid: Not sure what's the best approach to also add in support for `{http://jats.nlm.nih.gov}title`. Can see one of

- Separate visitor (explicit, more code that does exactly the same)
- Allow multiple possible elements here (explicit and reuses same code)
- Don't expect a specific element, like this (might behave strangely if given an element that isn't title like); means no code changes to support `{http://jats.nlm.nih.gov}trans-title` etc

vendor-extra/JatsContentBundle/src/Handler/JatsContentHandler.php
thewilkybarkid: I have wondered whether these handlers should be replaced by emitting an event, and listeners then populate bits of the page (works with the page grid having a set of 'slots' that they use /cc @davidcmoulton).

thewilkybarkid: Both are HMVC-like, but that would be more separated (the single handler here being replaced by a number of event listeners).

config/services.yaml
```diff
@@ -1,3 +1,6 @@
+parameters:
+    locale: 'en'
```
giorgiosironi: how do people configure this (if it needs to be configured)? Should it have an environment variable that can be used?

giorgiosironi: (used to provide this value to the container)

vendor-extra/JatsContentBundle/src/Resources/config/services.xml
```diff
@@ -14,12 +14,36 @@
...
+            <tag name="libero.view_converter"/>
+        </service>
+
+        <service id="Libero\JatsContentBundle\ViewConverter\KeywordGroupTagListVisitor"
+            class="Libero\JatsContentBundle\ViewConverter\KeywordGroupTagListVisitor">
+            <argument type="service" id="Libero\ViewsBundle\Views\ViewConverter"/>
+            <argument type="service" id="translator"/>
+            <argument type="collection">
+                <!-- Map of @kwd-group-type values to translation keys -->
+                <argument key="author-keywords">libero.jats.keyword_group.title.author_keywords</argument>
```
giorgiosironi: are these names (`author-keywords`, `research-organism`) generic or are they supposed to be overridden via configuration?

vendor-extra/JatsContentBundle/src/ViewConverter/FrontItemTagsVisitor.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        $this->converter = $converter;
+    }
+
+    protected function doVisit(Element $object, View $view, array &$context = []) : View
+    {
+        /** @var Document $document */
+        $document = $object->ownerDocument;
+        $xpath = $document->xpath();
+        $xpath->registerNamespace('jats', 'http://jats.nlm.nih.gov');
```
giorgiosironi: sounds like a constant

vendor-extra/JatsContentBundle/src/ViewConverter/HeadingVisitor.php
giorgiosironi: Configurable set of elements for a single Visitor?

giorgiosironi:
I followed https://github.com/libero/browser/#getting-started to run this, but I think that's outdated because it's using the `LiberoContentHandler` rather than `JatsContentHandler`:
```
app_1  | [07-Feb-2019 14:19:44] WARNING: [pool www] child 6 said into stderr: "[2019-02-07 14:19:44] request.CRITICAL: Uncaught PHP Exception UnexpectedValueException: "Co
uld not find a front" at /app/vendor-extra/LiberoContentBundle/src/Handler/LiberoContentHandler.php line 35 {"exception":"[object] (UnexpectedValueException(code: 0): Coul
d not find a front at /app/vendor-extra/LiberoContentBundle/src/Handler/LiberoContentHandler.php:35)"} []"
```

thewilkybarkid:
@giorgiosironi Need to sort out the README and split `prod` vs `dev` (for the latter, don't create `config/packages/content_page.yaml`).

vendor-extra/JatsContentBundle/src/Resources/config/services.xml
thewilkybarkid: Default ones that Libero understands (part of the higher-level question, _what_ JATS are we supporting directly); can provide configuration to add in custom ones (doable as it stands, but requires code).

vendor-extra/JatsContentBundle/src/ViewConverter/FrontItemTagsVisitor.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Libero\JatsContentBundle\ViewConverter;
+
+use DOMNodeList;
```
nlisgo: My IDE recommended that we add `ext-dom` to `composer.json`.

thewilkybarkid: And it is correct.

(I'd like to enable [ComposerRequireChecker](https://github.com/maglnet/ComposerRequireChecker), but have seen false positives when trying it before. Looks like there's config options now so might be doable.)

PR #37, opened 2019-02-08
Replace content handlers with events

thewilkybarkid:
Based on https://github.com/libero/browser/pull/35#discussion_r254614771; removes the individual content handlers and instead uses events. Should be more flexible to add bits to the page, relies on having slots available to add content too. (At the minute just have 'content'.)


phpstan.neon.dist
```diff
@@ -1,5 +1,6 @@
 parameters:
     ignoreErrors:
+        - '~^Access to an undefined property FluentDOM\\DOM\\Xpath::\$registerNodeNamespaces\.$~'
```
thewilkybarkid: https://github.com/ThomasWeinert/FluentDOM/commit/799210204cec9c4dc7a87bf168df63307e17d269

vendor-extra/ContentPageBundle/src/Event/CreateContentPageEvent.php
```diff
@@ -0,0 +1,59 @@
...
+
+use FluentDOM\DOM\Document;
+use Symfony\Component\EventDispatcher\Event;
+
+class CreateContentPageEvent extends Event
+{
+    public const NAME = 'libero.page.content';
+
+    private $content = [];
+    private $context;
```
thewilkybarkid: A lot of these things could be extracted into a more generic page event (eg search page, listing page).

vendor-extra/ContentPageBundle/src/Resources/config/services.xml
```diff
@@ -8,6 +8,13 @@
 
         <defaults public="false"/>
 
+        <service id="libero.content_page.page.namespace"
+            class="Libero\ContentPageBundle\EventListener\NamespaceListener">
+            <argument>libero</argument>
+            <argument>http://libero.pub</argument>
```
thewilkybarkid: This is a bit annoying, but trying to avoid having to redeclare namespaces everywhere.

README.md
```diff
@@ -16,11 +16,9 @@ To run a website reading from two content services (`blog-articles` and `scholar
     content_page:
         pages:
             blog_article:
-                handler: 'libero'
```
thewilkybarkid: I'm wondering whether the [`<meta>` of an item](https://github.com/libero/schemas/pull/14) could declare what type of content it is (currently 'libero' or 'jats', but could be something custom). Reading this would allow us to narrow down what's triggered, without having to query the content at run time.

giorgiosironi: What does "querying the content at run time" entails? You can envision situations when multiple formats for the content are present; maybe it's a matter of content negotiation if we can define a `Content-Type` more precise than `application/xml`.

vendor-extra/ContentPageBundle/src/Event/CreateContentPageEvent.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public function getContent() : array
+    {
+        return $this->content;
+    }
+
+    public function addContent($content) : void
+    {
+        $this->content[] = $content;
```
giorgiosironi: this mutability stands in constrant with ordinary usage of immutable events, I'm interested in possible ways of separating the propagation of the event from a collecting the `content` being produced by listeners. Still wouldn't want a global service collecting everything as that has an even larger scope of mutability.

vendor-extra/ContentPageBundle/src/Controller/ContentController.php
```diff
@@ -48,17 +49,20 @@ public function __invoke(Request $request, string $id) : Response
...
                     ];
 
+                    $event = new CreateContentPageEvent($document, $context);
+                    $this->dispatcher->dispatch($event::NAME, $event);
+
                     return new Response(
                         $this->twig->render(
                             $this->template,
-                            $this->contentHandler->handle($dom->documentElement, $context)
+                            $event->getContext() + ['title' => $event->getTitle(), 'content' => $event->getContent()]
```
giorgiosironi: for reference: `title` is only `<title>`, but not any h1 or similar content in the page, I think

giorgiosironi: It seems like the order of elements in `content` is important, because they will be rendered in order in the HTML page by `text.html.twig`

thewilkybarkid: Yep, order of the listener execution is important, so the content header ones are high up https://github.com/libero/browser/blob/a9b3a68c4af2876d2a0e4d8d843d24a66703fbb2/vendor-extra/JatsContentBundle/src/Resources/config/services.xml#L14

vendor-extra/ContentPageBundle/src/Event/CreateContentPageEvent.php
thewilkybarkid: Symfony events are really hooks.

README.md
thewilkybarkid: > What does "querying the content at run time" entails?
 https://github.com/libero/browser/blob/a9b3a68c4af2876d2a0e4d8d843d24a66703fbb2/vendor-extra/JatsContentBundle/src/EventListener/ContentHeaderListener.php#L22-L28

So it always listens but only cares about JATS content. This will be a problem if the content contains both Libero XML _and_ JATS (possible?). If it were declared somewhere that Libero is the primary format in that case, listeners like this could be excluded.

thewilkybarkid:
The advantage would be adding new bits on to the page. The handlers did all of that (eg content header here, body text here, reference list there). Adding in a new widget or whatever (eLife's metric graphs, for example) somewhere on that page isn't possible without duplicating the handler.

I did wonder whether this could replace the visitors as well, but they are doing different things. The view converters are turning a data structure into patterns ('give me the pattern for this paragraph'), these listeners are determining where they go on the page ('put the body content in the middle').

> events which are meant to propagate data only in one direction

[Bad naming upstream.](https://blog.ircmaxell.com/2012/03/handling-plugins-in-php.html#Observer-Pattern-1)

giorgiosironi:
Still the documentation at https://symfony.com/doc/current/components/event_dispatcher.html doesn't sanction mutable events or bidirectional communication like in a general Mediator (the dispatcher) that is injected into all collaborating objects? The only provided way to provoke side-effects from a subscriber is a limited "$e->stopPropagation()" call which is similar to what the DOM does. I suppose the collection of data could be generating other events, if the dispatcher is made available in the right places; but it seems more oriented to provide extension points into an existing flow rather than implementing the whole response generation process with events.

thewilkybarkid:
> implementing the whole response generation process with events

[Symfony's HttpKernel does this.](https://symfony.com/doc/current/components/http_kernel.html#httpkernel-driven-by-events)

> seems more oriented to provide extension points into an existing flow rather than implementing the whole response generation process with events

Yeah, the goal is to add an extension point, and it made sense to dog-food (and more flexibility too, as you could turn them off/replace them). It is _another_ layer of complexity though. Will have to think about creating a page in the Symfony profiler that can tell you what actually happened when constructing the page.

giorgiosironi:
Fits with the framework then - [the events used](https://symfony.com/doc/current/components/http_kernel.html#component-http-kernel-event-table) seem to have various accumulators for data like the `Request` bag.

README.md
giorgiosironi: Should work as long as there is a clear priority between the two different listeners (otherwise one would overwrite/clash with the other?)

vendor-extra/ContentPageBundle/src/Controller/ContentController.php
```diff
@@ -48,17 +49,20 @@ public function __invoke(Request $request, string $id) : Response
...
                 function (ResponseInterface $response) use ($request) : Response {
-                    $dom = FluentDOM::load((string) $response->getBody());
+                    $document = FluentDOM::load((string) $response->getBody());
 
                     $context = [
                         'lang' => $request->getLocale(),
                         'dir' => text_direction($request->getLocale()),
                     ];
 
+                    $event = new CreateContentPageEvent($document, $context);
```
giorgiosironi: naming matches the example https://github.com/symfony/symfony/tree/4.2/src/Symfony/Component/HttpKernel/Event

vendor-extra/ContentPageBundle/src/Event/CreateContentPageEvent.php
giorgiosironi: (when we have >=2 pages)

vendor-extra/ContentPageBundle/src/Resources/config/services.xml
giorgiosironi: It could be done by the controller when creating the event object rather than being modelled as mutable? Would still need configuration probably, but it may lead to a Factory for the events when we have more than one.

vendor-extra/JatsContentBundle/src/EventListener/ContentHeaderListener.php
```diff
@@ -0,0 +1,40 @@
...
+        $this->converter = $converter;
+    }
+
+    public function onCreatePage(CreateContentPageEvent $event) : void
+    {
+        $xpath = $event->getItem()->xpath();
+
+        $front = $xpath->firstOf('/libero:item/jats:article/jats:front');
+
+        if (!$front instanceof Element) {
```
giorgiosironi: should these checks be positive, e.g. `$front === null`?

vendor-extra/JatsContentBundle/src/EventListener/ContentHeaderListener.php
```diff
@@ -0,0 +1,40 @@
...
+            return;
+        }
+
+        $title = $xpath->firstOf('jats:article-meta/jats:title-group/jats:article-title', $front);
+
+        if ($title instanceof Element) {
+            $event->setTitle((string) $title);
```
giorgiosironi: in the current model, the last listener wins as `setTitle()` calls always override an already present title; should this method raise an exception if called multiple times?

thewilkybarkid: This case should check it's not already set. There might be reasons to decorate it.

vendor-extra/JatsContentBundle/src/EventListener/ContentHeaderListener.php
thewilkybarkid: This helps static analysis (and maybe humans) understand the type, since XPath can return many different types.

PR #38, opened 2019-02-08
Libero italic should match schema

nlisgo:
See: https://github.com/libero/schemas/blob/master/extensions/inline/italic.rng#L51-L65

### DIFF
--- a/vendor-extra/LiberoContentBundle/tests/ViewConverter/ItalicVisitorTest.php
+++ b/vendor-extra/LiberoContentBundle/tests/ViewConverter/ItalicVisitorTest.php
@@ -37,8 +37,8 @@ final class ItalicVisitorTest extends TestCase
 
     public function nodeProvider() : iterable
     {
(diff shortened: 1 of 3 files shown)


vendor-extra/LiberoContentBundle/tests/ViewConverter/ItalicVisitorTest.php
```diff
@@ -37,8 +37,8 @@ public function it_does_nothing_if_it_is_not_a_libero_italic_element(string $xml
 
     public function nodeProvider() : iterable
     {
-        yield 'different namespace' => ['<i xmlns="http://example.com">foo</i>'];
-        yield 'different element' => ['<b xmlns="http://libero.pub">foo</b>'];
+        yield 'different namespace' => ['<italic xmlns="http://example.com">foo</italic>'];
+        yield 'different element' => ['<bold xmlns="http://libero.pub">foo</bold>'];
```
nlisgo: Decided to change these also in anticipation on support for `libero:bold`. See: https://github.com/libero/schemas/pull/15

PR #42, opened 2019-02-13
Add support for sub in JatsContentBundle

nlisgo:
Improves: https://github.com/libero/libero/issues/106

Blocked by: https://github.com/libero/pattern-library/pull/51


nlisgo:
@thewilkybarkid you mentioned that we can hold off on advancing the libero schema for now. Does that mean that we don't need to add a `SubVisitor` to the `LiberoContentBundle` at this moment?

thewilkybarkid:
@nlisgo Could you try out #44? The Sub pattern came first, so we can avoid having the Sub update here.

vendor-extra/JatsContentBundle/src/Resources/config/services.xml
```diff
@@ -100,6 +100,12 @@
             <tag name="libero.view_converter"/>
         </service>
 
+        <service id="Libero\JatsContentBundle\ViewConverter\SubVisitor"
+                 class="Libero\JatsContentBundle\ViewConverter\SubVisitor">
```
thewilkybarkid:
```suggestion
            class="Libero\JatsContentBundle\ViewConverter\SubVisitor">
```

vendor-extra/JatsContentBundle/tests/ViewConverter/SubVisitorTest.php
```diff
@@ -0,0 +1,116 @@
...
+     * @test
+     */
+    public function it_sets_the_template_and_text_argument() : void
+    {
+        $visitor = new SubVisitor($this->createDumpingConverter());
+
+        $element = $this->loadElement(
+            <<<XML
+<jats:sub xmlns:jats="http://jats.nlm.nih.gov">
+    foo <jats:bold>bar</jats:bold> baz
```
thewilkybarkid: While we're not allowing mixed content in the spec, right to leave it working here (as we're allowing it to be changed).

PR #44, opened 2019-02-13
Allow the pattern update script to fetch pull requests

thewilkybarkid:
Should help cases where there's been multiple merges on the Pattern Library and we need incremental PRs here, as well as experimental usage (ie the Pattern Library PR _hasn't_ yet been merged, eg for the likes of #42).

/cc @nlisgo


bin/update-patterns
```diff
@@ -9,6 +9,24 @@ function finish {
     docker rm browser-pattern-library &> /dev/null || true
 }
 
+if [[ $# -eq 0 ]]
```
giorgiosironi: I'd pass this script through `shellcheck`, which will probably make you add a lot of `"` around variables for safety and other suggestions

bin/update-patterns
```diff
@@ -9,6 +9,24 @@ function finish {
...
+  then
```
giorgiosironi: `then` usually on the same line as `if`?

bin/update-patterns
thewilkybarkid: Already done!

bin/update-patterns
```diff
@@ -9,6 +9,21 @@ function finish {
     docker rm browser-pattern-library &> /dev/null || true
 }
 
+if [[ $# -eq 0 ]]; then
+    commit=origin/master
+elif [[ $# -eq 1 && $1 =~ ^pr-([1-9][0-9])*$ ]]; then
+    pr=${BASH_REMATCH[1]}
+elif [[ $# -eq 1 ]]; then
+    commit=$1
+else
```
nlisgo: How can I trigger the `else` condition?

thewilkybarkid: More than 1 argument.

PR #47, opened 2019-02-26
Add the grid

thewilkybarkid:

PR #50, opened 2019-03-18
Store the context in the View

thewilkybarkid:
Following https://github.com/libero/browser/pull/47#discussion_r266006504.

Didn't originally do this at the `View` is meant for output, but it did meaning having to pass around `$context` by reference. Could have used a mutable object instead, but #47 is looking like there's use cases for knowing the resulting context elsewhere, like https://github.com/libero/browser/blob/e216d4bc3a50dbeaa62c2f3e7433fff5b8d9e985/vendor-extra/ContentPageBundle/src/Event/CreateContentPagePartEvent.php#L42

### DIFF
--- a/vendor-extra/ViewsBundle/src/Views/View.php
+++ b/vendor-extra/ViewsBundle/src/Views/View.php
@@ -6,16 +6,19 @@ namespace Libero\ViewsBundle\Views;
 
 use JsonSerializable;
 use function array_replace_recursive;
+use function is_string;
 
 final class View implements JsonSerializable
 {
     private $arguments;
+    private $context;
     private $template;
 
-    public function __construct(?string $template, array $arguments = [])
+    public function __construct(?string $template, array $arguments = [], array $context = [])
     {
         $this->template = $template;
         $this->arguments = $arguments;
+        $this->context = $context;
     }
 
     public function hasArgument(string $key) : bool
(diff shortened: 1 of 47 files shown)


vendor-extra/ViewsBundle/src/Views/View.php
```diff
@@ -69,6 +86,15 @@ public function withTemplate(string $template) : View
         return $view;
     }
 
+    public function withContext(array $context) : View
+    {
+        $view = clone $this;
+
+        $view->context = array_replace_recursive($view->context, $context);
+
```
giorgiosironi: guarantees immutability

vendor-extra/ViewsBundle/tests/Views/ViewConverterRegistryTest.php
```diff
@@ -67,28 +67,38 @@ public function it_delegates_to_visitors() : void
 
         $handler->add(
             new CallbackVisitor(
-                function (Element $object, View $view, array &$context = []) : View {
-                    return $view->withTemplate($view->getTemplate().'foo')->withArgument('foo', 'foo');
+                function (Element $object, View $view) : View {
+                    return $view
+                        ->withTemplate($view->getTemplate().'foo')
+                        ->withArgument('foo', 'foo')
+                        ->withContext(['one' => 'foo']);
```
giorgiosironi: isn't it weird to replace everything inside a view? Should return a new one?

thewilkybarkid: Only `withTemplate` overwrites, the other two are merged in.

giorgiosironi: (potential naming inconsistency? It's probably a base type used often enough that usage will become clear anyway)

PR #51, opened 2019-03-19
[WIP] Teaser list and homepage

nlisgo:


vendor-extra/ContentPageBundle/src/Controller/ContentController.php
```diff
@@ -36,37 +39,88 @@ public function __construct(
...
+                                );
+                        }, iterator_to_array($document->xpath()->evaluate('libero:item-ref')));
+
+                        $event = new CreateHomePageEvent($list, $context);
+                        $this->dispatcher->dispatch($event::NAME, $event);
+
+                        return new Response(
+                            $this->twig->render(
+                                $this->template,
+                                $event->getContext() + ['title' => 'Home', 'content' => $event->getContent()]
```
nlisgo: Justing setting the title manually for now will use the approach we have taken for some headings and aria-labels.

vendor-extra/ContentPageBundle/src/Controller/ContentController.php
```diff
@@ -36,37 +39,88 @@ public function __construct(
...
+                        $list = array_map(function (Element $element) use ($request, $context) {
+                            return $this->client
+                                ->requestAsync(
+                                'GET',
+                                "{$element->getAttribute('service')}/items/{$element->getAttribute('id')}/versions/latest"
+                                )
+                                ->then(
+                                    function (ResponseInterface $response) use ($request, $context) : Document {
+                                        return FluentDOM::load((string) $response->getBody());
+                                    }
```
nlisgo: As each item in the list could return libero or JATS XML do we need to trigger an event for each item response?

.docker/api/data/scholarly-articles/article2/1.xml
```diff
@@ -0,0 +1,153 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nlisgo: Just adding another item so we can have a list

vendor-extra/ContentPageBundle/src/Controller/ContentController.php
thewilkybarkid: Something like https://github.com/libero/browser/blob/master/vendor-extra/JatsContentBundle/src/EventListener/BodyListener.php should pick this up, which will cycle through the items and convert them to teasers. The handling of that part will be view converts in the `JatsContentBundle` or `LiberoContentBundle`. However, they'll be the full item rather than just the relevant `<front>`, so might need to go through more than one...

nlisgo: If I understand you correctly, we can get away with triggering a single event for the list and then have a listener for the whole list within `LiberoContentBundle` but the conversion of each item into views will be delegated to the content bundle which recognises the XML namespace of that item.

PR #52, opened 2019-03-19
Add homepage

thewilkybarkid:


thewilkybarkid:
Experiment here is to replace `ContentPageBundle` with a more generic `LiberoPageBundle`. This means extracting out how content is fetched (`LoadPageEvent`) then turned into patterns (`CreatePageEvent`). All types of pages can then use the simple `PageController`.

vendor-extra/JatsContentBundle/src/EventListener/BodyListener.php
```diff
@@ -21,9 +21,13 @@ public function __construct(ViewConverter $converter)
         $this->converter = $converter;
     }
 
-    public function onCreatePage(CreateContentPageEvent $event) : void
+    public function onCreatePage(CreatePageEvent $event) : void
     {
-        $xpath = $event->getItem()->xpath();
+        if ('content' !== $event->getRequest()->attributes->get('libero_page')['type']) {
```
thewilkybarkid: It'd be useful to make this so listeners aren't called for things they don't care about.

vendor-extra/LiberoPageBundle/src/DependencyInjection/LiberoPageConfiguration.php
```diff
@@ -45,12 +45,27 @@ private function getPagesDefinition() : ArrayNodeDefinition
...
+                    ->ifTrue(function (array $values) : bool {
+                        return 'content' === $values['type'] && !isset($values['content_service']);
+                    })
+                    ->thenInvalid('Content pages require a content_service')
+                ->end()
+                ->validate()
+                    ->ifTrue(function (array $values) : bool {
+                        return 'content' !== $values['type'] && isset($values['content_service']);
+                    })
+                    ->thenInvalid('Non-content pages cannot have a content_service')
```
thewilkybarkid: This is a bit messy. Don't know if there's a better way to have a discriminator value (`type`), but maybe the different types should have their own configuration instead.

vendor-extra/LiberoPageBundle/src/Event/LoadPageEvent.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public function addDocument(string $key, PromiseInterface $promise) : void
+    {
+        $this->documents[$key] = $promise;
+    }
+
+    public function getDocuments() : PromiseInterface
+    {
+        return all($this->documents);
```
thewilkybarkid: This will be problematic when handling lists: there will be follow-up requests for each item, which won't be included here.

vendor-extra/LiberoPageBundle/src/EventListener/TitleListener.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        $this->translator = $translator;
+    }
+
+    public function onCreatePage(CreatePageEvent $event) : void
+    {
+        $title = $event->getTitle();
+
+        if (is_string($title)) {
+            $title .= ' | ';
```
thewilkybarkid: Not sure if needs localisation?

vendor-extra/LiberoPageBundle/src/Event/LoadPageEvent.php
thewilkybarkid: Wonder if they can be the same event? 🤔

vendor-extra/LiberoPageBundle/src/EventListener/LiberoPageListener.php
```diff
@@ -0,0 +1,40 @@
...
+            return;
+        }
+
+        $page = $this->routes[$route];
+
+        if ('content' === $page['type']) {
+            $page['content_id'] = $request->attributes->get('id');
+        }
+
+        $request->attributes->set('libero_page', $page);
```
thewilkybarkid: This is the key part. Attaches the details (page type etc) about the Libero page to the request (based on the matched route).

vendor-extra/LiberoPageBundle/src/Event/LoadPageEvent.php
thewilkybarkid:
```php
return all($this->documents)->then(
    function () : PromiseInterface {
        return all($this->documents);
    }
);
```

looks weird, but might work?

thewilkybarkid: Thinking about this, we probably don't want to be eager-loading everything. This would result in a lot of chained requests (eg list > item > item categories > category). HTTP caching helps, but avoiding the requests entirely would be better (ie caching the thing that would be produced, like the teaser).

vendor-extra/LiberoPageBundle/src/EventListener/TitleListener.php
nlisgo: It's been a while since I have worked on a site with multiple languages but we will need instances of translatable text which accept variables such as `Welcome back, [username]`. Can we have a translatable string for the site title when page and site title is available: `[page_title] | [site_title]`.

thewilkybarkid: 👍

vendor-extra/LiberoPageBundle/src/Resources/translations/messages+intl-icu.en.xlf
```diff
@@ -0,0 +1,23 @@
...
+
+    <file source-language="en" datatype="plaintext" original="file.ext">
+
+        <body>
+
+            <trans-unit id="1" resname="libero.page.site_name">
+                <source>Site Name</source>
+            </trans-unit>
+
+            <trans-unit id="2" resname="libero.page.page_title">
```
thewilkybarkid: Maybe should be a more specific name? It's for the `<title>` in the HTML `<head>`, so maybe `libero.page.head.title`?

vendor-extra/LiberoPageBundle/src/Event/LoadPageEvent.php
thewilkybarkid: Done a quick experiment and seems to work. Essentially needs replacing `View` with one that wraps the HTTP request and conversion. https://github.com/libero/browser/blob/5d8b471cab69b20b93997291ef41d46722eb22e1/vendor-extra/ViewsBundle/src/Views/ViewConverterRegistry.php#L36-L38 needs consideration as accessing details should unwrap it, so need a way to stop propagation (special context key?).

vendor-extra/LiberoPageBundle/src/DependencyInjection/LiberoPageConfiguration.php
thewilkybarkid: Maybe better to group them by type, so something like:

```yaml
libero_page:
    pages:
        homepage:
            name: 'homepage'
            path: '/'
        content:
            - name: blog_article
              path: '/blog/{id}'
              content_service: 'blog-articles'
            - name: scholarly_article:
              path: '/articles/{id}'
              content_service: 'scholarly-articles'
```

So each group can either be a single entry, or a list of entries.

vendor-extra/LiberoPageBundle/src/Event/LoadPageEvent.php
thewilkybarkid: This is demoed in #56.

PR #54, opened 2019-03-21
Try and make the README a bit clearer

thewilkybarkid:


.env.dist
```diff
@@ -1,3 +1,4 @@
+API_URI=http://libero-api
```
giorgiosironi: but this is not read in https://github.com/libero/browser/blob/master/docker-compose.yaml#L15?

README.md
```diff
@@ -8,9 +8,24 @@ Serve content read from a Libero API to the public.
...
+1. Run `docker-compose down --volumes --remove-orphans && docker-compose up --build`.
 
-1. Create `config/packages/content_page.yaml`:
+2. Open http://localhost:8080/articles/article1 to see the scholarly article `article1` in the `dev` environment.
+
+### Configuration
+
+The `dev` environment reads content from a [Libero dummy API](https://github.com/libero/dummy-api), using the configuration and data in [`.docker/api`](.docker/api).
+
+Running in production
```
giorgiosironi: inconsistent title format `---` vs `###`?

thewilkybarkid: There's underline styles only for levels 1 and 2.

.env.dist
thewilkybarkid: Was thinking this is a form of documentation about what can be set generally (ie not just for the local Docker Compose setup), but maybe that's best kept elsewhere.

README.md
```diff
@@ -8,9 +8,24 @@ Serve content read from a Libero API to the public.
...
+---------------------
+
+The application is not yet stable, but Docker images are published ([`liberoadmin/browser`](https://hub.docker.com/r/liberoadmin/browser)).
+
+To run an image reading from two content services (`blog-articles` and `scholarly-articles`), you will need the following configuration:
+
+- Set the `API_URI` environment variable to be the root of the Libero API.
```
giorgiosironi: If we are abstracting this as "the Libero API", does it matter that there are two content services behind it? Can we still talk of individual services if all we see from here is a gateway?

README.md
giorgiosironi: But why? :thinking:

README.md
thewilkybarkid: The API spec is [all about services](https://docs.google.com/document/d/1JhVpW8dxAuJqnk2YsQa76yxdIV_BlOd1QIi02fIu9mo/edit#heading=h.ule1npii5d5y), rather than individual endpoints.

giorgiosironi: A bit overloaded as a term but ok

thewilkybarkid: I'm all up for improving names. 😃

PR #56, opened 2019-03-26
Add homepage with list of teasers

thewilkybarkid:
Experimental PR that won't be merged. Contains #52, plus how loading complete lists could work.

PR #57, opened 2019-03-27
Replace ViewConverterRegistry with EventDispatchingViewConverter

thewilkybarkid:
Based on the need in #56 to be able to stop event propagation. Get integration logging/profiling in Symfony for free (refs https://github.com/libero/browser/pull/56#discussion_r269467618).

Given Symfony is also moving to adopt the new [PSR-14](https://www.php-fig.org/psr/psr-14/) (https://github.com/symfony/symfony/pull/30691), we won't be tying ourselves to the framework.


vendor-extra/ViewsBundle/src/Event/CreateViewEvent.php
```diff
@@ -0,0 +1,38 @@
...
+
+declare(strict_types=1);
+
+namespace Libero\ViewsBundle\Event;
+
+use FluentDOM\DOM\Element;
+use Libero\ViewsBundle\Views\View;
+use Symfony\Component\EventDispatcher\Event;
+
+final class CreateViewEvent extends Event
```
thewilkybarkid: Needs a better name.

vendor-extra/JatsContentBundle/src/EventListener/CreateView/FrontSubjectGroupContentHeaderListener.php
```diff
@@ -2,37 +2,37 @@
...
+    use SimplifiedViewConverterListener;
 
     public function __construct(ViewConverter $converter, TranslatorInterface $translator)
     {
         $this->converter = $converter;
         $this->translator = $translator;
     }
 
-    protected function doVisit(Element $object, View $view) : View
+    protected function handle(CreateViewEvent $event) : View
```
thewilkybarkid: Easier to pass the `Element` and `View` directly, but won't be able to stop events. Maybe the `Event` should be a third argument, and (mostly) not needed?

thewilkybarkid: Works, but PHPStan (correctly) complains `invoked with 3 parameters, 2 required`. Unless it is typehinted and left unused.

thewilkybarkid: Maybe if you want the event it shouldn't be using the 'simplified' trait. The template etc validation could be extracted into a separate one.

thewilkybarkid: dabae39

vendor-extra/ViewsBundle/src/Views/SimplifiedViewConverterListener.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Libero\ViewsBundle\Views;
+
+use FluentDOM\DOM\Element;
+use Libero\ViewsBundle\Event\CreateViewEvent;
+
+trait SimplifiedViewConverterListener
```
thewilkybarkid: A lot of the changes caused by this have been extracted into #58.

giorgiosironi:
Seems to conflict with https://github.com/libero/browser/pull/58#pullrequestreview-219943252 now?

composer.lock
```diff
@@ -2958,7 +2959,7 @@
                     "homepage": "https://nette.org/contributors"
                 }
             ],
-            "description": "🅱 Nette Bootstrap: the simple way to configure and bootstrap your Nette application.",
+            "description": "? Nette Bootstrap: the simple way to configure and bootstrap your Nette application.",
```
thewilkybarkid: 😒

thewilkybarkid:
@giorgiosironi Updated now.

vendor-extra/ViewsBundle/src/Event/CreateViewEvent.php
thewilkybarkid: `ConvertView`?

vendor-extra/JatsContentBundle/tests/EventListener/CreateView/FrontArticleTitleContentHeaderListenerTest.php
```diff
@@ -100,11 +107,13 @@ public function it_does_nothing_if_there_is_already_a_content_title_set() : void
 </front>
 XML
         );
-        ;
```
thewilkybarkid: Related: https://github.com/libero/php-coding-standard/pull/55

vendor-extra/ViewsBundle/src/Event/CreateViewEvent.php
thewilkybarkid: `GetViewForElement`? `ElementToView`?

vendor-extra/JatsContentBundle/src/EventListener/CreateView/SupListener.php
```diff
@@ -2,20 +2,19 @@
...
 use FluentDOM\DOM\Element;
 use Libero\ViewsBundle\Views\ConvertsChildren;
-use Libero\ViewsBundle\Views\OptionalTemplateVisitor;
+use Libero\ViewsBundle\Views\OptionalTemplateListener;
 use Libero\ViewsBundle\Views\View;
 use Libero\ViewsBundle\Views\ViewConverter;
-use Libero\ViewsBundle\Views\ViewConverterVisitor;
 use function Libero\ViewsBundle\array_has_key;
 
-final class SupVisitor implements ViewConverterVisitor
```
giorgiosironi: there's no interface anymore? It's probably inherited from the traits but worth to express it here if present?

thewilkybarkid: The equivalent is to implement [`EventSubscriberInterface`](https://github.com/symfony/symfony/blob/4.1/src/Symfony/Component/EventDispatcher/EventSubscriberInterface.php) rather than declaring it the service definition https://github.com/libero/browser/blob/e7e06056df2403657dd404c99684e3365f7fe19e/vendor-extra/JatsContentBundle/src/Resources/config/services.xml#L118 (which means it's not coupled to the Symfony dispatcher)

vendor-extra/ViewsBundle/src/Event/CreateViewEvent.php
giorgiosironi: Verbs are usually out of place wrt nouns; we also don't have a system metaphor (a real world system that we could borrow naming from). `ViewCreationFromElement`?

vendor-extra/JatsContentBundle/src/EventListener/CreateView/SupListener.php
thewilkybarkid: Could have

```php
interface CreateViewListener
{
    public function onCreateView(CreateViewEvent $event) : void;
}
```

but it wouldn't be required.

vendor-extra/ViewsBundle/src/Event/CreateViewEvent.php
giorgiosironi: It doesn't create the view because it is passed in the constructor, though it may replace it later?

thewilkybarkid: Correct. Other options are:

- Pass in the arguments for the `View` and create it there.
- Manipulate the arguments for the `View` so it's only created on `getView()`(/`toView()`). Means this event is mutable, which fits as it's actually the mediator pattern.

giorgiosironi: Is this a [`ViewBuilder`](https://www.oodesign.com/builder-pattern.html) then?

thewilkybarkid: 👍  `BuildViewEvent`?

giorgiosironi: (something that indicates it's a `Builder` pattern, if it is so)

thewilkybarkid:
Travis build not starting, fine on my machine so :shipit:.

PR #58, opened 2019-03-27
Combine simplified visitors

thewilkybarkid:
Extracted from #57.


vendor-extra/ViewsBundle/src/Views/SimplifiedVisitor.php
```diff
@@ -5,14 +5,20 @@
...
+use LogicException;
 use function in_array;
+use function is_string;
 use function sprintf;
 
 trait SimplifiedVisitor
 {
     final public function visit(Element $object, View $view) : View
     {
-        if ($this->expectedTemplate() !== $view->getTemplate()) {
```
giorgiosironi: Tried to review from mobile and just didn't work.

Difference between `expectedTemplate` and `possibleTemplate` is unclear, can they be documented with docblocks here? There is no separate interface for them so this is the best place, also it's where all the dependencies fan in.

thewilkybarkid: Might need to rethink this a little. Currently `expectedTemplate` is for cases where the `View` is requesting a specific template (eg ContentHeader), whereas `possibleTemplate` is for cases where the `View` might not know what template it needs (eg when cycling through body content, which contains paragraphs, sections, lists etc).

thewilkybarkid: Had a go in 23fdf4a.

giorgiosironi: This ~class~ trait now makes few assumptions about its children.

PR #60, opened 2019-03-28
Enable Xdebug in dev

thewilkybarkid:
Surprisingly easy to get working with PhpStorm on a Mac (https://blog.jetbrains.com/phpstorm/2018/08/quickstart-with-docker-in-phpstorm/).


Dockerfile
```diff
@@ -106,6 +106,16 @@ ENV APP_ENV=dev
 USER root
 ENV COMPOSER_ALLOW_SUPERUSER=true
 
+RUN apk add --no-cache --virtual .build-deps $PHPIZE_DEPS && \
+    pecl install \
```
giorgiosironi: I don't like compiling things for a dynamic language, but no alternative

.docker/php-dev.ini
```diff
@@ -1 +1,4 @@
 opcache.validate_timestamps = 1
+
+xdebug.remote_enable = 1
+xdebug.remote_host = host.docker.internal
```
giorgiosironi: This is the host where an IDE runs, I think. Might be worth making `xdebug.remote_port` explicit rather than relying on the default.

Dockerfile
thewilkybarkid: Did wonder if the compilation could be done separately and the result copied in to save app changes from re-running it.

giorgiosironi: In theory the end result is just a `.so` file in the right place, which if the PHP (and OS?) version match should be movable across filesystems. Maintaining the code to move it around might not be worth it though if the primary use case of `dev` is overriding the code with a host folder mounted as a volume.

thewilkybarkid: Will ignore for now, can revisit if it does become annoying.

thewilkybarkid:
Just tried it out a bit more, and it's _very_ slow on Docker for Mac. https://jtreminio.com/blog/developing-at-full-speed-with-xdebug/ looks an interesting technique.

.docker/nginx/browser.conf
```diff
@@ -14,7 +21,8 @@ server {
         include fastcgi_params;
         internal;
 
-        fastcgi_pass app:9000;
+        fastcgi_pass $fastcgi_pass:9000;
+        fastcgi_read_timeout 1d; # For Xdebug
```
thewilkybarkid: Seems like variables can't be set with `time` values.

docker-compose.override.yaml
```diff
@@ -8,3 +8,22 @@ services:
...
+        environment:
+            API_URI: http://web:8080
+            APP_SECRET:
+            DEFAULT_LOCALE: ${DEFAULT_LOCALE:-en}
+        volumes:
+            - ./:/app
+            - /app/var
+            - /app/vendor
+        depends_on:
+            - api
```
thewilkybarkid: Unfortunately `extends` was removed from Docker Compose.

giorgiosironi: So this is basically a copy of `app` with a derived image, which lives only locally. Fair.

.docker/nginx/browser.conf
```diff
@@ -1,3 +1,10 @@
+resolver 127.0.0.11;
+
+map $cookie_XDEBUG_SESSION $fastcgi_pass {
```
giorgiosironi: if this needs to set something specific in the IDE/host, it would need documentation

.docker/nginx/browser.conf
```diff
@@ -1,3 +1,10 @@
+resolver 127.0.0.11;
```
giorgiosironi: is this a standard for `docker` or `docker-compose`?

thewilkybarkid: Docker.

.docker/nginx/browser.conf
thewilkybarkid: Should work with any value.

PR #61, opened 2019-03-29
Add translations directory

thewilkybarkid:
Translations can already be mounted in, this makes it a bit more explicit.

Also overrides the path in `test`, as https://github.com/libero/browser/blob/1bc13dc32ccf32480a15cdd97b73a35c15c043a5/tests/WebTest.php#L25 will be testing it.

### DIFF
--- a/config/packages/test/translation.yaml
+++ b/config/packages/test/translation.yaml
@@ -1,0 +1,3 @@
+framework:
+    translator:
+        default_path: '%kernel.project_dir%/tests/translations'
--- a/config/packages/test/translation.yaml
+++ b/config/packages/test/translation.yaml
@@ -1,3 +1,3 @@
 framework:
     translator:
-        default_path: '%kernel.project_dir%/tests/translations'
+        default_path: '/var/empty'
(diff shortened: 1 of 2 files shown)


config/packages/test/translation.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+    translator:
+        default_path: '%kernel.project_dir%/tests/translations'
```
giorgiosironi: does this folder actually exist?

thewilkybarkid: No, `/dev/null` is probably better.

thewilkybarkid: `file_exists('/dev/null')` is `true`. 🤔

giorgiosironi: It exists (on Linux), but it's not a regular file:
```
$ test -f /dev/null
$ echo $?
1
$ test -e /dev/null                  
$ echo $?                            
0
$ ls -l /dev/null 
crw-rw-rw- 1 root root 1, 3 Mar 29 08:47 /dev/null
```

PR #63, opened 2019-04-02
Separate View into different implementations

thewilkybarkid:
#56 gave rise to the idea of having a lazy `View`. That experiment hacked it in; but it's better to have a real proxy implementation. After a bit of thought, we _already_ have the need for different implementations (primarily to stop simple strings having to be rendered as a template). So, after a bit of a refactor in https://github.com/libero/pattern-library/pull/76 we can have a separate `StringView` to handle that case.


thewilkybarkid:
As pointed out by @nlisgo, doesn't actually rely on https://github.com/libero/pattern-library/pull/76 so we can separate them.

vendor-extra/ViewsBundle/src/Views/TemplateView.php
```diff
@@ -0,0 +1,93 @@
...
+    public function withContext(array $context) : TemplateView
+    {
+        $view = clone $this;
+
+        $view->context = array_replace_recursive($view->context, $context);
+
+        return $view;
+    }
+
+    public function jsonSerialize()
```
giorgiosironi: this is not serializing to JSON (a string conforming to https://json.org/); it's serializing to a PHP data structure that can be encoded in JSON if needed (associative array, string, null)

giorgiosironi: But it's a PHP standard so imposed

giorgiosironi: Why are we serializing to JSON? Only debugging comes to mind, as an equivalent of `var_dump()`/`var_export()`

thewilkybarkid: Yeah, it was useful when debugging. Since Xdebug has been set up, won't be needed.

thewilkybarkid:
Bit of refactoring done with #64 in mind.

nlisgo:
Happy with these changes. It is a useful distinction.

thewilkybarkid:
Any more comments @giorgiosironi?

PR #64, opened 2019-04-02
Add list of content to the homepage

nlisgo:


vendor-extra/LiberoPageBundle/src/EventListener/BuildView/ItemListListener.php
```diff
@@ -0,0 +1,58 @@
...
+                    );
+                },
+                iterator_to_array($object->getElementsByTagNameNS('http://libero.pub', 'item-ref'))
+            )
+        );
+    }
+
+    protected function canHandleTemplate(string $template) : bool
+    {
+        return '@LiberoPatterns/text.html.twig' === $template;
```
nlisgo: I've had issues changing this to `teaser-list`. I will submit some followup commits.

vendor-extra/JatsContentBundle/src/EventListener/BuildView/ItemTeaserListener.php
```diff
@@ -0,0 +1,61 @@
...
+                $object
+            );
+
+        if (!$heading instanceof Element) {
+            return $view;
+        }
+
+        return $view
+            // @todo - set route to content.
+            ->withArgument('href', '#')
```
nlisgo: We will need to access the routes here.

nlisgo:
Since https://github.com/libero/browser/pull/64/commits/84a422845e268b542a265bdc40dd9c7dfce83aea I am experiencing an issue with the `heading` variable not being available in the teaser template. I suspect the href is not available either. Not sure why this small adjustment is causing the teaser template not to work.

vendor-extra/LiberoPageBundle/src/EventListener/BuildView/ItemListListener.php
```diff
@@ -0,0 +1,62 @@
...
+            'list',
+            [
+                'items' => array_map(
+                    function (NonDocumentTypeChildNode $child) use ($view) : array {
+                        return [
+                            'content' => $this->converter->convert(
+                                $child,
+                                '@LiberoPatterns/teaser.html.twig',
+                                $view->getContext()
+                            )
```
thewilkybarkid: Ref https://github.com/libero/browser/pull/64#issuecomment-479404044:

```suggestion
                            )->getArguments()
```

The converter returns the complete `View`, but the list template already includes the inclusion of the `teaser` template https://github.com/libero/browser/blob/029a20136526946ce13e2a7b01549368191ea4f7/vendor-extra/LiberoPatternsBundle/src/Resources/views/teaser-list.html.twig#L37 so the `teaser` template was ending up with the `View` rather than the arguments for it.

vendor-extra/JatsContentBundle/src/EventListener/BuildView/ItemTeaserListener.php
thewilkybarkid: This will be something similar to https://github.com/elifesciences/journal/blob/develop/src/Router/ModelParameterResolver.php (rather than 'create this route for me', it's 'give me the path for this piece of content').

Will have a think.

vendor-extra/LiberoPageBundle/src/DependencyInjection/LiberoPageConfiguration.php
```diff
@@ -72,14 +75,12 @@ private function getContentPagesDefinition() : ArrayNodeDefinition
...
         $pagesNode
+            ->normalizeKeys(false)
             ->arrayPrototype()
                 ->children()
                     ->scalarNode('path')
                         ->isRequired()
                     ->end()
-                    ->scalarNode('content_service')
-                        ->isRequired()
-                    ->end()
```
thewilkybarkid: This change should be extracted. This assumes there's 1-to-1 between a content service and a content page (can't believe you would duplicate them?), and removes the name in favour of the service id. Using that in the route allows us to generate the route name from elsewhere, like https://github.com/libero/browser/blob/137582fa1f021ede7ed22c6075178eab604ede9a/vendor-extra/LiberoPageBundle/src/EventListener/BuildView/ItemTeaserHrefListener.php#L39

thewilkybarkid: https://github.com/libero/browser/pull/65

nlisgo:
I'm resolving the conflicts now.

vendor-extra/LiberoPageBundle/src/EventListener/BuildView/ItemRefTeaserListener.php
```diff
@@ -0,0 +1,55 @@
...
+                "{$object->getAttribute('service')}/items/{$object->getAttribute('id')}/versions/latest"
+            )
+            ->then(
+                function (ResponseInterface $response) use ($object, $view) : View {
+                    $item = FluentDOM::load((string) $response->getBody());
+                    $item->namespaces($object->ownerDocument->namespaces());
+
+                    return $this->converter->convert($item->documentElement, $view->getTemplate(), $view->getContext());
+                }
+            )->wait();
```
nlisgo: We need to introduce `LazyView`. @thewilkybarkid can we pair on this tomorrow briefly. I have your code sample but I think the master branch has diverged from that fairly significantly.

vendor-extra/JatsContentBundle/src/EventListener/BuildView/ItemArticleTitleTeaserListener.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    protected function canHandleTemplate(?string $template) : bool
+    {
+        return '@LiberoPatterns/teaser.html.twig' === $template;
+    }
+
+    protected function canHandleElement(string $element) : bool
+    {
+        return '{http://libero.pub}item' === $element;
```
nlisgo: Should this be `return '{http://jats.nlm.nih.gov}front' === $element;`?

config/packages/dev/libero_page.yaml
```diff
@@ -2,6 +2,7 @@ libero_page:
     pages:
         homepage:
             path: '/'
+            search_service: 'scholarly-articles'
```
thewilkybarkid: Been thinking about how to power the list on the homepage. eLife uses its search service, which is where our initial direction has been here. But that has been a bit problematic as it's not then easy for us to turn off search if needed (it'd be good to have a circuit breaker there). For Libero though, this needs to be a list of items. As it stands, the content API and search API lists are the same so can be interchanged. Search will be expanded to allow filtering by type etc. If we are able to keep the basic definition of how lists work the same, then it's possible to use other things here (eg could be a dedicated list). Means the config here would be a path, as opposed to the name of a service. (To use a search service, would be the path include any query string parameters, which as it stands has to be configured somewhere else.)

/cc @giorgiosironi

vendor-extra/LiberoPageBundle/tests/TwigTestCase.php
```diff
@@ -21,7 +22,7 @@ trait TwigTestCase
         $twig->method('render')
             ->willReturnCallback(
                 function (...$arguments) : string {
-                    return '<html><body>'.json_encode($arguments).'</body></html>';
+                    return '<html><body>'.json_encode(dump($arguments)).'</body></html>';
```
thewilkybarkid: Already extracted into https://github.com/libero/browser/pull/68.

config/packages/dev/libero_page.yaml
thewilkybarkid: Done in 30d382dabb2c158610c5bf59ac918e41e579c47b. There's already a mismatch in the dervied paths (a search service would be just `/{service-name}`, whereas as content API is `/{service-name}/items`).

config/packages/dev/libero_page.yaml
```diff
@@ -2,6 +2,7 @@ libero_page:
     pages:
         homepage:
             path: '/'
+            primary_listing: 'scholarly-articles/items'
```
giorgiosironi: seems flexible enough to allow the different services, does it need additions to `sample-configuration` then?

vendor-extra/JatsContentBundle/tests/EventListener/BuildView/FrontArticleTitleTeaserListenerTest.php
```diff
@@ -0,0 +1,170 @@
...
+    </article-meta>
+</front>
+XML
+        );
+
+        $event = new BuildViewEvent($element, new TemplateView('template'));
+        $listener->onBuildView($event);
+        $view = $event->getView();
+
+        $this->assertInstanceOf(TemplateView::class, $view);
```
giorgiosironi: I understand the negative cases but what would be a way of shortening the assertion phases of these tests? Extract a custom assertion method that contains this phase seems cheap and effective.

vendor-extra/JatsContentBundle/tests/EventListener/BuildView/FrontArticleTitleTeaserListenerTest.php
```diff
@@ -0,0 +1,170 @@
...
+        $this->assertSame('@LiberoPatterns/teaser.html.twig', $view->getTemplate());
+        $this->assertSame(['heading' => 'bar'], $view->getArguments());
+        $this->assertEmpty($view->getContext());
+    }
+
+    /**
+     * @test
+     */
+    public function it_sets_the_text_argument() : void
```
giorgiosironi: the title says `text_argument` but the test does something else?

vendor-extra/LiberoContentBundle/tests/EventListener/BuildView/FrontTitleTeaserListenerTest.php
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * @test
+     */
+    public function it_sets_the_text_argument() : void
+    {
+        $listener = new FrontTitleTeaserListener($this->createDumpingConverter());
+
+        $element = $this->loadElement(
+            <<<XML
+<libero:front xmlns:libero="http://libero.pub">
```
giorgiosironi: a pity that this is a completely duplicated test in another bundle, that just changes the input from JATS to `libero` XML.

vendor-extra/LiberoPageBundle/src/EventListener/BuildView/ItemListListener.php
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    protected function handle(Element $object, TemplateView $view) : View
+    {
+        /** @var DOMNodeList<Element> $itemRefs */
+        $itemRefs = $object('libero:item-ref');
+
+        $level = $view->getContext()['level'] ?? 1;
+        if ($view->hasContext('list_title')) {
+            $level++;
```
giorgiosironi: mutable counter in a long method, tricky

config/packages/dev/libero_page.yaml
thewilkybarkid: https://github.com/libero/browser/blob/3bb139182d3978a26d32ea23e2deffe2bb895585/vendor-extra/LiberoPageBundle/src/DependencyInjection/LiberoPageConfiguration.php#L64-L66 Is required, but can point at either content or search API (anything that returns a `libero:item-list`).

vendor-extra/LiberoContentBundle/tests/EventListener/BuildView/FrontTitleTeaserListenerTest.php
thewilkybarkid: Might be able to do something to help these different implementations and their test suites, but definitely one for the future.

vendor-extra/JatsContentBundle/tests/EventListener/BuildView/FrontArticleTitleTeaserListenerTest.php
thewilkybarkid: `this->assertEquals(new TemplateView('template'), $view);` works.

vendor-extra/LiberoPageBundle/src/EventListener/BuildView/ItemListListener.php
```diff
@@ -0,0 +1,110 @@
...
+        return new LazyView(
+            function () use ($view, $items) {
+                return $view->withArgument(
+                    'list',
+                    [
+                        'items' => array_filter(
+                            array_map(
+                                function (View $view) {
+                                    if (!$view instanceof ArrayAccess) {
+                                        return [];
```
giorgiosironi: what does this case cover? It looks like we are skipping an item, but don't know why we would have an empty element here

vendor-extra/LiberoPageBundle/src/EventListener/BuildView/ItemRefTeaserHrefListener.php
```diff
@@ -0,0 +1,56 @@
...
+        /** @var string $service */
+        $service = $object->getAttribute('service');
+
+        if ('' === $id || '' === $service) {
+            return $view;
+        }
+
+        return $view->withArgument(
+            'href',
+            $this->urlGenerator->generate("libero.page.content.{$service}", ['id' => $id])
```
giorgiosironi: good abstraction of UrlGenerator as an application service that is passed in where needed

vendor-extra/LiberoPageBundle/src/EventListener/BuildView/ItemRefTeaserListener.php
```diff
@@ -0,0 +1,83 @@
...
+use Psr\Http\Message\ResponseInterface;
+
+final class ItemRefTeaserListener
+{
+    use SimplifiedViewConverterListener;
+
+    private $client;
+    private $converter;
+
+    public function __construct(ClientInterface $client, ViewConverter $converter)
```
giorgiosironi: mixing the `Listeners` with `Clients` is the dangerous part in terms of layering, but I'd agree there's no other clear option here at the moment

vendor-extra/JatsContentBundle/tests/EventListener/BuildView/FrontArticleTitleTeaserListenerTest.php
thewilkybarkid: Exists in other tests too, will follow up with a refactor.

vendor-extra/LiberoPageBundle/src/EventListener/BuildView/ItemListListener.php
thewilkybarkid: Can move the check higher up. `ViewConverter::convert` could return a `StringView`. Might want to throw an exception though, or at least log it.

PR #66, opened 2019-04-04
Add article publication date

thewilkybarkid:


vendor-extra/JatsContentBundle/src/EventListener/BuildView/IsoDateListener.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    private const DATE_PATTERN = '~^([0-9]{4}-[0-9]{2}-[0-9]{2})(?:$|T)~';
+
+    use SimplifiedViewConverterListener;
+
+    protected function handle(Element $object, View $view) : View
+    {
+        $date = $object->getAttribute('iso-8601-date');
+
+        if (!is_string($date)) {
```
thewilkybarkid: Would use `hasAttribute()` here, but PHPStan doesn't then recognise that `getAttribute()` will only return a string afterwards.

PR #69, opened 2019-04-10
Use listing grid on the homepage

thewilkybarkid:
Implements https://github.com/libero/pattern-library/pull/78.


vendor-extra/LiberoPatternsBundle/src/constants.php
```diff
@@ -8,8 +8,13 @@
 const PAGE_GRID_MAIN = 'main';
 const PAGE_GRID_END = 'end';
 
-const CONTENT_GRID_PRIMARY = 'primary';
+const MAIN_GRID_FULL = 'full';
+const MAIN_GRID_MAIN = 'main';
+const MAIN_GRID_PRIMARY = 'primary';
```
thewilkybarkid: Based on the assumption that _all_ 'main' grids have `full`, `main`, and `primary` areas. /cc @davidcmoulton

vendor-extra/LiberoPageBundle/src/EventListener/MainListener.php
```diff
@@ -24,11 +24,20 @@ public function __construct(EventDispatcherInterface $dispatcher)
 
     public function onCreatePage(CreatePageEvent $event) : void
     {
+        switch ($event->getRequest()->attributes->get('libero_page')['type'] ?? null) {
+            case 'homepage':
+                $grid = '@LiberoPatterns/listing-grid.html.twig';
+                break;
+            default:
+                $grid = '@LiberoPatterns/content-grid.html.twig';
```
thewilkybarkid: Might want the default to be a simple 1-column grid? /cc @davidcmoulton

vendor-extra/LiberoPageBundle/src/EventListener/MainListener.php
```diff
@@ -24,11 +24,20 @@ public function __construct(EventDispatcherInterface $dispatcher)
 
     public function onCreatePage(CreatePageEvent $event) : void
     {
+        switch ($event->getRequest()->attributes->get('libero_page')['type'] ?? null) {
+            case 'homepage':
```
giorgiosironi: special case doesn't feel very extension, but if the alternative is some sort of complex configuration that maps `homepage` to `listing-grid`...

thewilkybarkid: Did wonder about that, but not sure how exchangeable they are. The template itself can be overridden, which is probably enough for now.

vendor-extra/LiberoPatternsBundle/src/Resources/views/listing-grid.html.twig
```diff
@@ -0,0 +1,50 @@
+{%- set attributes = attributes|default({})|merge({class: attributes.class|default([])|merge(['listing-grid']) }) -%}
```
giorgiosironi: is this from an update to `pattern-library`? How is the version of `pattern-library` tracked? (I would assume it's not because there is no diff in `composer.lock` or similar lock file)

thewilkybarkid: It's not at the moment, result of running `bin/update-patterns pr-78`.

PR #70, opened 2019-04-10
Fix return type in SectionListener

thewilkybarkid:
Not currently visible in tests as https://github.com/libero/browser/blob/5bfb50be30fe26382daf6a8eb690556287c1470d/vendor-extra/LiberoPageBundle/tests/ViewConvertingTestCase.php#L17-L27 returns a `TemplateView`. Makes me think we need a specialist type there.

### DIFF
--- a/vendor-extra/JatsContentBundle/src/EventListener/BuildView/SectionListener.php
+++ b/vendor-extra/JatsContentBundle/src/EventListener/BuildView/SectionListener.php
@@ -10,6 +10,7 @@ use FluentDOM\DOM\Node\NonDocumentTypeChildNode;
 use Libero\ViewsBundle\Views\ConvertsChildren;
 use Libero\ViewsBundle\Views\OptionalTemplateListener;
 use Libero\ViewsBundle\Views\TemplateView;
+use Libero\ViewsBundle\Views\View;
 use Libero\ViewsBundle\Views\ViewConverter;
 use function array_map;
 use function iterator_to_array;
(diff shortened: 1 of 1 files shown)


giorgiosironi:
`createDumpingConverter` could specify a stricter type in the anonymous function it uses, but currently it specifies an interface (`View`) rather than a concrete class instead which is a good thing. Little difference.

thewilkybarkid:
Similar to https://github.com/libero/browser/pull/68 and https://github.com/libero/browser/pull/64#discussion_r273452172; might better to allow the tests to compare a normalized (JSON-like) form. So `ViewConvertingTestCase::createDumpingConverter()` would return a special (anonymous?) `View` implementation and requires using `dump()` to check it. (JSON-like so is similar to how Twig will read it.)

vendor-extra/JatsContentBundle/src/EventListener/BuildView/SectionListener.php
```diff
@@ -54,7 +55,7 @@ protected function handle(Element $object, TemplateView $view) : TemplateView
         return $view->withArgument(
             'content',
             array_map(
-                function (NonDocumentTypeChildNode $child) use ($childContext) : TemplateView {
+                function (NonDocumentTypeChildNode $child) use ($childContext) : View {
```
giorgiosironi: https://github.com/libero/browser/blob/master/vendor-extra/ViewsBundle/src/Views/CallbackViewConverter.php#L19 returns `View` so this matches that.

PR #71, opened 2019-04-12
Separate choosing a template from building a view

thewilkybarkid:
Allows `TemplateView` to always take a template and avoid https://github.com/libero/browser/blob/f02935fc996e5df6d497ad206d658fbf710b8488/vendor-extra/ViewsBundle/src/Views/OptionalTemplateListener.php#L20-L23 (which feels a bit of a hack).


vendor-extra/JatsContentBundle/tests/EventListener/BodyListenerTest.php
```diff
@@ -102,7 +102,7 @@ public function it_adds_the_body_content(string $xml, array $context, array $exp
         $this->assertEquals(
             array_map(
                 function (array $block) : TemplateView {
-                    return new TemplateView(null, $block);
+                    return new TemplateView('', $block);
```
thewilkybarkid: This is a bit of a pain; becoming clear that https://github.com/libero/browser/blob/f02935fc996e5df6d497ad206d658fbf710b8488/vendor-extra/LiberoPageBundle/tests/ViewConvertingTestCase.php#L17-L27 needs refactoring, but avoiding doing so here.

thewilkybarkid:
Can go through it in person, but in case it helps: this if for cases where we need to cycle through a list nodes rather than specifying a pattern for an individual element (eg going through body content, inline content). Currently setting a template happens at the same point as populating it with data. This means having a template has to be optional; this splits it up into two, so that’s no longer the case.

I’m trying to get to a point where we can dispatch an event for an individual pattern, rather than a general one. Means that the number of listeners is dramatically reduced, and should be more performant.

vendor-extra/JatsContentBundle/tests/EventListener/BuildView/FigGraphicFigureImageListenerTest.php
```diff
@@ -18,6 +19,29 @@ final class FigGraphicFigureImageListenerTest extends TestCase
     use ViewConvertingTestCase;
     use XmlTestCase;
 
+    /**
+     * @test
+     * @dataProvider templateChoiceProvider
+     */
+    public function it_can_choose_a_template(string $xml, ?string $expected) : void
+    {
+        $listener = new FigGraphicFigureListener($this->createFailingConverter());
```
giorgiosironi: would expect this to be the test of `FigGraphicFigureImageListener`, departing from convention to organize tests by use case?

vendor-extra/JatsContentBundle/src/EventListener/BuildView/FigGraphicFigureListener.php
```diff
@@ -7,15 +7,17 @@
...
+use Libero\ViewsBundle\Views\ViewBuildingListener;
 use Libero\ViewsBundle\Views\ViewConverter;
 use function Libero\ViewsBundle\array_has_key;
 
 final class FigGraphicFigureListener
 {
     use ConvertsChildren;
-    use OptionalTemplateListener;
+    use TemplateChoosingListener;
+    use ViewBuildingListener;
```
giorgiosironi: I was checking this as an example, and it does both the template choosing and the view building. Now looking for a case where only one happens.

vendor-extra/JatsContentBundle/src/EventListener/BuildView/FigCaptionTitleFigureListener.php
```diff
@@ -6,14 +6,14 @@
...
-use Libero\ViewsBundle\Views\OptionalTemplateListener;
 use Libero\ViewsBundle\Views\TemplateView;
+use Libero\ViewsBundle\Views\ViewBuildingListener;
 use Libero\ViewsBundle\Views\ViewConverter;
 
 final class FigCaptionTitleFigureListener
 {
     use ConvertsChildren;
-    use OptionalTemplateListener;
+    use ViewBuildingListener;
```
giorgiosironi: also this one (I think) not a case of separation of template and view, as I can see it produces a `TemplateView` with `@LiberoPatterns/figure.html.twig` at the template

giorgiosironi: Input and output of this Listener:
```
<jats:title>Dinaledi skeletal specimens.</jats:title>'
```
```
Libero\ViewsBundle\Views\TemplateView::__set_state(array(
   'arguments' =>  
  array (
    'level' => 3,
    'text' =>  
    array (
      0 =>  
      Libero\ViewsBundle\Views\StringView::__set_state(array(
         'string' => 'Dinaledi skeletal specimens.',
         'context' =>  
        array (
          'area' => 'primary',
          'level' => 3,
          'lang' => 'en',
          'dir' => 'ltr',
        ),  
      )), 
    ),  
  ),  
   'template' => '@LiberoPatterns/heading.html.twig',
   'context' =>                                                                                                                                                            
  array (
    'area' => 'primary',
    'level' => 3,
    'lang' => 'en',
    'dir' => 'ltr',
  ),  
))
```

vendor-extra/ViewsBundle/src/Views/StringViewConverter.php
```diff
@@ -0,0 +1,29 @@
...
+declare(strict_types=1);
+
+namespace Libero\ViewsBundle\Views;
+
+use FluentDOM\DOM\CdataSection;
+use FluentDOM\DOM\Element;
+use FluentDOM\DOM\Node\NonDocumentTypeChildNode;
+use FluentDOM\DOM\Text;
+
+final class StringViewConverter implements ViewConverter
```
giorgiosironi: I guess this is a key part

giorgiosironi: There is also `EmptyViewConverter` that produces `View` objects that do not refer to a template; but it's only used in tests.

vendor-extra/ViewsBundle/tests/Views/StringViewConverterTest.php
```diff
@@ -0,0 +1,79 @@
...
+        $this->assertInstanceOf(ViewConverter::class, $converter);
+    }
+
+    /**
+     * @test
+     * @dataProvider visibleNodeProvider
+     */
+    public function it_returns_a_string_view_for_visible_nodes(string $node, string $expected) : void
+    {
+        $converter = new StringViewConverter($this->createFailingConverter());
```
giorgiosironi: from the test it isn't clear whether the `fallback` inner converter is not called or whether its failure is caught; I'd substitute it with a mock that specified `$this->never()` if that's the intention

vendor-extra/ViewsBundle/tests/Views/StringViewConverterTest.php
```diff
@@ -0,0 +1,79 @@
...
+
+        $this->assertEquals(
+            new StringView($expected, ['bar' => 'baz']),
+            $converter->convert($this->loadNode($node), 'template', ['bar' => 'baz'])
+        );
+    }
+
+    public function visibleNodeProvider() : iterable
+    {
+        yield 'cdata' => ['<![CDATA[<cdata>]]>', '<cdata>'];
```
giorgiosironi: always confusing to work with `CDATA`, could the content be `42 > 41` rather than `<cdata>` so that it is generic character data rather than another "escaped" element which is a specific use case?

vendor-extra/JatsContentBundle/tests/EventListener/BuildView/FigGraphicFigureImageListenerTest.php
thewilkybarkid: Good spot.

vendor-extra/ViewsBundle/tests/Views/StringViewConverterTest.php
thewilkybarkid: Looks like the relevant test was broken, fixed in f1b0bdede0c2cf113e04346c447128b5ebfb4b64.

PR #77, opened 2019-04-23
Add figure titles/captions

thewilkybarkid:


vendor-extra/JatsContentBundle/src/EventListener/BuildView/FigCaptionContentFigureListener.php
```diff
@@ -0,0 +1,62 @@
...
+            },
+            iterator_to_array($paragraphs)
+        );
+
+        return $view->withArgument('caption', $caption);
+    }
+
+    protected function template() : string
+    {
+        return '@LiberoPatterns/figure.html.twig';
```
giorgiosironi: Weird that the `figure.html.twig` template is specified in two different `template()` methods from two different Listeners. It shouldn't be a problem if it's only used for matching, but it ends up https://github.com/libero/browser/blob/ef47d36fad5c45bc37ff6fc4770605c7759a24cf/vendor-extra/ViewsBundle/src/Views/OptionalTemplateListener.php#L22 so there's the potential for conflicts between the two e.g. forgetting to update one of the pair.

thewilkybarkid: https://github.com/libero/browser/blob/f02935fc996e5df6d497ad206d658fbf710b8488/vendor-extra/ViewsBundle/src/Views/SimplifiedViewConverterListener.php#L18-L20 and https://github.com/libero/browser/blob/f02935fc996e5df6d497ad206d658fbf710b8488/vendor-extra/ViewsBundle/src/Views/OptionalTemplateListener.php#L15-L18 means it won't be changed on subsequent ones (they'd just be skipped if different). But this is where https://github.com/libero/browser/pull/71 would help by making choosing the template clearer.

PR #79, opened 2019-05-01
Update fonts

thewilkybarkid:


templates/page.html.twig
```diff
@@ -12,7 +12,8 @@
 
     <meta name="format-detection" content="telephone=no">
 
-    <link href="{{ asset('bundles/liberopatterns/css/all.css') }}" rel="stylesheet">
+    <link href="{{ asset('bundles/liberopatterns/css/fonts.css') }}" rel="stylesheet">
```
giorgiosironi: I guess this allows independent caching of the two, as fonts change less often

thewilkybarkid: Yep, but main benefit is being able to replace the fonts entirely (if you need a different subset) without having to rebuild all the main styles.

PR #84, opened 2019-05-28
Update FluentDOM to 7.2

thewilkybarkid:


vendor-extra/JatsContentBundle/src/EventListener/BuildView/BoldListener.php
```diff
@@ -33,9 +33,9 @@ protected function template() : string
         return '@LiberoPatterns/bold.html.twig';
     }
 
-    protected function canHandleElement(string $element) : bool
+    protected function canHandleElement(Element $element) : bool
     {
-        return '{http://jats.nlm.nih.gov}bold' === $element;
+        return '{http://jats.nlm.nih.gov}bold' === $element->clarkNotation();
```
giorgiosironi: they could have used a `__toString()` method

thewilkybarkid: Nah, `__toString()` is commonly misused like that. `__toString()` is meant to be able to turn the object into a string, but can be turned back into an object without any data loss. Can't be done on a DOM element due to its location in the DOM etc. (FluentDOM _does_ implement `__toString()`, but returns the node text content. Which is also wrong.)

PR #87, opened 2019-06-13
Add authors and affiliations to the Content Header

thewilkybarkid:


vendor-extra/JatsContentBundle/src/EventListener/BuildView/ContribLinkListener.php
```diff
@@ -0,0 +1,104 @@
...
+            $order,
+            function (array $parts, string $component) use ($name, $view, $xpath) : array {
+                $element = $xpath->firstOf($component, $name);
+
+                if (!$element instanceof Element) {
+                    return $parts;
+                }
+
+                if (0 !== count($parts)) {
+                    $parts[] = ' ';
```
thewilkybarkid: This is annoying.

thewilkybarkid: And is actually wrong for some languages (eg Korean).

vendor-extra/ViewsBundle/src/Views/ConvertsChildren.php
```diff
@@ -16,6 +17,10 @@ trait ConvertsChildren
 
     final protected function convertChildren(Element $object, array $context = []) : array
     {
+        if ($context['strip_inline'] ?? false) {
```
thewilkybarkid: Not sure whether we should do this or not.

vendor-extra/JatsContentBundle/src/EventListener/BuildView/FrontContribAuthorAffContentHeaderListener.php
```diff
@@ -0,0 +1,80 @@
...
+use Libero\ViewsBundle\Views\ViewBuildingListener;
+use Libero\ViewsBundle\Views\ViewConverter;
+use function array_reduce;
+use function array_unique;
+use function array_values;
+use function count;
+use function Libero\ViewsBundle\array_has_key;
+use const SORT_REGULAR;
+
+final class FrontContribAuthorAffContentHeaderListener
```
giorgiosironi: at first I thought `Aff` was an abbreviation but it's an actual element

vendor-extra/JatsContentBundle/tests/EventListener/BuildView/FrontContribAuthorAffContentHeaderListenerTest.php
```diff
@@ -0,0 +1,232 @@
...
+
+    /**
+     * @test
+     */
+    public function it_sets_the_affiliations_argument() : void
+    {
+        $listener = new FrontContribAuthorAffContentHeaderListener(
+            new CallbackViewConverter(
+                static function (NonDocumentTypeChildNode $node, ?string $template = null, array $context = []) : View {
+                    if (!$node instanceof Element
```
giorgiosironi: extracting a single type of element as `text` seems like a common use case, is this `CallbackViewConverter` something that should be created by a superclass, to make it available to many tests?

giorgiosironi: (similar to `createDumpingConverter`)

thewilkybarkid: Complexity here is due to requiring `text` (which is incompatible with `createDumpingConverter()` ) https://github.com/libero/browser/blob/ac23393ee2459a529e28534daea90a96597c587c/vendor-extra/JatsContentBundle/src/EventListener/BuildView/FrontContribAuthorAffContentHeaderListener.php#L48-L49

Though maybe it should fail rather than skip in that case?

vendor-extra/ViewsBundle/src/Views/ConvertsChildren.php
thewilkybarkid: This is the implementation of https://github.com/libero/jats-support/pull/13#discussion_r294214068.

vendor-extra/JatsContentBundle/tests/EventListener/BuildView/FrontContribAuthorAffContentHeaderListenerTest.php
thewilkybarkid: Turns out the name fails if nothing sets the `text`, so should just match.

Could do with more useful error messages though.

thewilkybarkid: Which also relates to https://github.com/libero/jats-support/pull/13#discussion_r294242059.

thewilkybarkid: Just realised

| Complexity here is due to requiring `text`

isn't true, it's due to the `array_unique`...

thewilkybarkid: Should probably move this elsewhere.

thewilkybarkid: Wrong commit name, but done in 568a3a398807eeabe08a0b84fd161e643c45618c.

vendor-extra/ViewsBundle/src/Views/TemplateView.php
```diff
@@ -65,7 +65,7 @@ public function withContext(array $context) : TemplateView
     {
         $view = clone $this;
 
-        $view->context = array_replace_recursive($view->context, $context);
+        $view->context = array_replace($view->context, $context);
```
thewilkybarkid: `array_replace_recursive()` was causing problems as the list wasn't being replaced entirely.

vendor-extra/LiberoPageBundle/src/EventListener/BuildView/ContentHeaderAffiliationDuplicateListener.php
```diff
@@ -0,0 +1,48 @@
...
+    use ViewBuildingListener;
+
+    protected function handle(Element $object, TemplateView $view) : View
+    {
+        $affiliations = $view->getArgument('affiliations');
+
+        $affiliations['items'] = array_values(
+            array_unique(
+                $view->getArgument('affiliations')['items'] ?? [],
+                SORT_REGULAR
```
thewilkybarkid: This is a bit too simplistic (a string and a `StringTemplate` should be considered the same if their text is the same), but can be improved later.

vendor-extra/JatsContentBundle/src/EventListener/BuildView/LinkListener.php
```diff
@@ -38,8 +38,12 @@ protected function canHandleElement(Element $element) : bool
         return string_is(
             $element->clarkNotation(),
             '{http://jats.nlm.nih.gov}aff',
+            '{http://jats.nlm.nih.gov}given-names',
             '{http://jats.nlm.nih.gov}kwd',
-            '{http://jats.nlm.nih.gov}subject'
+            '{http://jats.nlm.nih.gov}prefix',
+            '{http://jats.nlm.nih.gov}subject',
+            '{http://jats.nlm.nih.gov}suffix',
+            '{http://jats.nlm.nih.gov}surname'
```
thewilkybarkid: This is a bit annoying. We don't really want a Link, just text (which might have a `lang` attribute etc). Might need a Text pattern that is a `<span>` or not depending if there's attributes.

thewilkybarkid:
Actually ready to go now.

vendor-extra/LiberoPageBundle/src/EventListener/BuildView/ContentHeaderAffiliationDuplicateListener.php
nlisgo: > This is a bit too simplistic (a string and a StringTemplate should be considered the same if their text is the same), but can be improved later.

Would it be appropriate to add a comment in the code or open an issue and link it here?

PR #92, opened 2019-06-26
Add info bar

davidcmoulton:
Implements https://github.com/libero/pattern-library/pull/115.

Needs some work. Aware it's conflating an infobar pattern's generic implementation with the use case of the demo journal site, so needs a layer of configuration to replace the hard coded parts. No tests as not sure yet which way this would go.


vendor-extra/LiberoPageBundle/src/EventListener/InfoBarListener.php
```diff
@@ -0,0 +1,45 @@
...
+use Libero\LiberoPageBundle\Event\CreatePagePartEvent;
+use Libero\ViewsBundle\Views\ContextAwareTranslation;
+use Libero\ViewsBundle\Views\TemplateView;
+use Symfony\Contracts\Translation\TranslatorInterface;
+use const Libero\LiberoPatternsBundle\MAIN_GRID_FULL;
+
+final class InfoBarListener
+{
+    use ContextAwareTranslation;
+    public const ATTENTION = ['name' => 'attention', 'imagePath' => 'bundles/liberopatterns/images/attention.svg'];
```
davidcmoulton: This way of specifying images is wrong, they will need proper asset handling.

vendor-extra/LiberoPageBundle/src/EventListener/InfoBarListener.php
```diff
@@ -21,10 +20,9 @@ final class InfoBarListener
 
     private $urlGenerator;
 
-    public function __construct(TranslatorInterface $translator, UrlGeneratorInterface $urlGenerator)
```
davidcmoulton: Having second thoughts about this, might be useful to generate the path to the image?

vendor-extra/LiberoPageBundle/src/Resources/translations/messages+intl-icu.en.xlf
```diff
@@ -24,6 +24,10 @@
                 <source>No items available.</source>
             </trans-unit>
 
+            <trans-unit id="5" resname="libero.page.info_bar.demo">
+                <source>This is a demo site.</source>
```
thewilkybarkid: Thinking we have a problem if we want to introduce a link...

thewilkybarkid: Currently HTML will be escaped. Either we find a hacky syntax here and process the results, or raw HTML is allowed (ideally in just specific cases).

thewilkybarkid: The latter is definitely simple (https://github.com/libero/browser/pull/92/commits/75af57ebb93aa867a8142b48a2b9e4464c191f19).

thewilkybarkid:
Going to postponed allowing inline HTML.

config/services.yaml
```diff
@@ -1,5 +1,6 @@
 parameters:
     env(DEFAULT_LOCALE): 'en'
+    env(INFO_BAR): null
```
thewilkybarkid: Not sure this (and the other) should be environment variables. More complicated to process, hence https://github.com/libero/browser/blob/a7b804f25e7dbfa58a2970720ba6ca55e00bebc3/vendor-extra/LiberoPageBundle/src/EventListener/InfoBarListener.php#L31-L33

Neither are things that should be changeable at runtime.

/cc @giorgiosironi

giorgiosironi: Do they change across environments? Environment variables can, but should not, be changed at runtime (as in "during execution rather than in the application bootstrap").

thewilkybarkid: They're resolved during the bootstrap for a request, but not the application itself (when the container is compiled).

giorgiosironi: So what's the alternative to environment variables, the configuration file? Seems to be similar in behavior.

thewilkybarkid: A config file, but it will only be read when compiling the container (ie needs a cache clear/redeploy to change).

thewilkybarkid: I'll keep it for now (to match the other), but probably something to review.

PR #94, opened 2019-07-02
Allow envvar to be nullable

thewilkybarkid:
Seems to get round https://github.com/libero/sample-configuration/pull/68#discussion_r299444241.


config/packages/libero_page_defaults.yaml
```diff
@@ -1,5 +1,5 @@
 libero_page:
     client: 'csa_guzzle.client.libero_api'
     info_bar:
-        text: '%env(INFO_BAR)%'
+        text: '%env(string:default::INFO_BAR)%'
```
thewilkybarkid: Wasn't expecting to have to use the `string` processor, but `ValidateEnvPlaceholdersPass` seems to get confused by the types that the [`default` processor can return](https://github.com/symfony/symfony/blob/5203393a84f9da80ef98417ea2dd05017426cf3e/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php#L47), and errors as https://github.com/libero/browser/blob/0857c9c46cdeec2ee0cfd81f03a4e3dbcb40a7e0/vendor-extra/LiberoPageBundle/src/DependencyInjection/LiberoPageConfiguration.php#L98 doesn't allow an `array`.

Appears _not_ to cast `null` (back) to an empty string though.

PR #95, opened 2019-07-04
Add references

thewilkybarkid:


thewilkybarkid:
We've hit a problem that I was worried about: _all_ the nodes going to _all_ the listeners (not counted, but that's going to be a big number) is too much for the profiler and runs out of memory. It still seems ok without (page rendering within ~150ms locally, no caching yet too), though it's pretty wasteful as most listeners won't do anything.

Will see if can exclude these events from the profiler, otherwise more granular events would work (1 event per template?).

vendor-extra/ViewsBundle/src/ViewsBundle.php
```diff
@@ -4,8 +4,14 @@
...
 
+use Libero\ViewsBundle\DependencyInjection\Compiler\DebugEventDispatcherPass;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
 final class ViewsBundle extends Bundle
 {
+    public function build(ContainerBuilder $container) : void
+    {
+        $container->addCompilerPass(new DebugEventDispatcherPass());
```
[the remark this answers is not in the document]
thewilkybarkid: Extracted https://github.com/libero/browser/pull/96

Dockerfile
```diff
@@ -85,6 +85,7 @@ USER root
 COPY --from=composer /usr/bin/composer /usr/bin/composer
 RUN touch .phpcs-cache && \
     chown www-data:www-data .phpcs-cache
+COPY .docker/php-test.ini ${PHP_INI_DIR}/conf.d/01-app.ini
```
thewilkybarkid: Same name as the `dev` config, so it's overwritten.

vendor-extra/LiberoPatternsBundle/src/Resources/views/reference.html.twig
```diff
@@ -0,0 +1,58 @@
...
+            {%- endif -%}
+
+          {%- endblock authors -%}
+        {%- endwith -%}
+
+        {%- with details|default({}) only -%}
+          {%- block details -%}
+
+            {%- if text is defined -%}
+              {%- include '@LiberoPatterns/paragraph.html.twig' -%}
```
thewilkybarkid: Not for now, but not sure of the best way to handle multiple pieces of information being brought together here.

vendor-extra/JatsContentBundle/tests/EventListener/ReferencesListenerTest.php
```diff
@@ -0,0 +1,230 @@
...
+        <jats:back>
+            <jats:ref-list/>
+            <jats:ref-list/>
+        </jats:back>
+    </jats:article>
+</libero:item>
+XML
+            ,
+            [
+                'lang' => 'en',
```
giorgiosironi: it seems the class under test is only passing down the context, so why testing with three examples of contexts that however have identical XML? This direction of variation doesn't seem special for this class

vendor-extra/JatsContentBundle/tests/EventListener/BuildView/RefListSectionListenerTest.php
```diff
@@ -0,0 +1,212 @@
...
+     * @test
+     */
+    public function it_uses_the_current_level() : void
+    {
+        $listener = new RefListSectionListener($this->createDumpingConverter(), new IdentityTranslator());
+
+        $element = $this->loadElement(
+            <<<XML
+<jats:ref-list xmlns:jats="http://jats.nlm.nih.gov">
+    <jats:title>Title</jats:title>
```
giorgiosironi: is this a realistic example? I thought there would be `<ref>` elements inside `<ref-list>`

giorgiosironi: looking at the expected values, know that elements inside here won't be used by this Listener, but would make the example complete (in documentation terms rather than coverage terms)

PR #96, opened 2019-07-05
Remove debug.event_dispatcher

thewilkybarkid:
Extracted from https://github.com/libero/browser/pull/95#discussion_r300619514.

The `dev` environment always has been a lot slower than `prod`, seems to be the profiling of the event dispatcher (generating the data for the patterns sees a _lot_ of events go to a _lot_ of listeners). Adding references sees it collapse entirely (runs out of memory).

This takes a brute-force approach and disables the special event dispatcher used for profiling. Means the 'Performance' profile pane is now empty.

Will try and think of a far better solution, but for now this works.


stephenwf:
Will this prevent debugging on local environments completely?

thewilkybarkid:
> Will this prevent debugging on local environments completely?

No, it's purely profiling of the event dispatcher.

PR #100, opened 2019-07-09
Fix test config order

thewilkybarkid:
Meant to put it _before_ code changes in #99, since it's far less likely to change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -85,8 +85,8 @@ USER root
 COPY --from=composer /usr/bin/composer /usr/bin/composer
 RUN touch .phpcs-cache && \
     chown www-data:www-data .phpcs-cache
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -85,8 +85,8 @@ USER root
 COPY --from=composer /usr/bin/composer /usr/bin/composer
 RUN touch .phpcs-cache && \
     chown www-data:www-data .phpcs-cache
-COPY tests/ tests/
 COPY .docker/php-test.ini ${PHP_INI_DIR}/conf.d/01-app.ini
+COPY tests/ tests/
```
giorgiosironi: (`COPY` operation are usually not expensive anyway in terms of build time)

PR #101, opened 2019-07-09
Add Google Tag Manager support

thewilkybarkid:


templates/page.html.twig
```diff
@@ -17,10 +17,31 @@
 
     <script defer src="{{ asset('js/main.bundle.js', 'libero_patterns') }}"></script>
 
+    {% if gtm_id %}
+        <script>
```
giorgiosironi: from here on, I'll trust it as a black box that has been copied and pasted

config/services.yaml
```diff
@@ -1,5 +1,6 @@
 parameters:
     env(DEFAULT_LOCALE): 'en'
+    env(GTM_ID): null
```
giorgiosironi: default seems good

PR #103, opened 2019-07-10
Add org.opencontainers.image.revision label

giorgiosironi:
Prerequisite for https://github.com/libero/environments/pull/47

Compare https://github.com/libero/search/pull/28


giorgiosironi:
For this use case other labels are not needed, but if we can decide on a standard set they could be consistently added across the projects. Not sure what the adoption level is...

PR #105, opened 2019-07-11
Add menu to the site header

thewilkybarkid:


vendor-extra/LiberoPageBundle/src/EventListener/SiteHeaderListener.php
```diff
@@ -37,6 +37,12 @@ public function onCreatePagePart(CreatePagePartEvent $event) : void
                             'alt' => $this->translate('libero.page.site_name', $context),
                         ],
                     ],
+                    'menu' => [
```
thewilkybarkid: This is clearly simplistic and not (easily) changeable, but the menu structure is likely to change in the near future.

